PR #2, opened 2016-09-19
handle conversion of label columns

sjhewitt:
when using sqlalchemy expressions as columns, the selectable is a Label which may be converted, however the Label type may not have `doc` or `nullable` attributes. This change uses `getattr` to try to retrieve them.

### DIFF
--- a/graphene_sqlalchemy/converter.py
+++ b/graphene_sqlalchemy/converter.py
@@ -87,7 +87,8 @@ def convert_sqlalchemy_type(type, column, registry=None):
 @convert_sqlalchemy_type.register(postgresql.ENUM)
 @convert_sqlalchemy_type.register(postgresql.UUID)
 def convert_column_to_string(type, column, registry=None):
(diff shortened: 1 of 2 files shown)


gsvitak:
I am trying to use a column_property with that maps to an Int. Based on the above change, I think the Int convert should be something like the following.

I have tested locally and it appears to work. Can you please let me know if I am on the right path? I can submit a pull request if think this is the right approach.

```
@convert_sqlalchemy_type.register(types.Integer)
def convert_column_to_int_or_id(type, column, registry=None):
    if column.primary_key:
        return ID(description=column.doc, required=not(column.nullable))
    else:
        return Int(description=getattr(column, 'doc', None),
                  required=not(getattr(column, 'nullable', True)))
```

sjhewitt:
that looks about right - I think we should probably use `getattr` in each of the converters

gsvitak:
Great!!. I already submitted a pull request for Int. Let me know if I can help with anything else.

PR #5, opened 2016-09-25
adding support to convert int label columns similar to PR#2

gsvitak:


gsvitak:
closing for pep formating issue causing Travis build to fail

PR #6, opened 2016-09-25
Convert int label column for SQLAlchemy column_property

gsvitak:


sjhewitt:
as mentioned in #2 , it would be great if the use of `getattr` was expanded to all of the converters in the file

PR #8, opened 2016-10-04
Set  iterable to returned connection of the connection_resolver.

varuna82:
Returned iterable with the connection similar to [IterableConnectionField](https://github.com/graphql-python/graphene/blob/c961f0b3c60c3a7a6223d5218c912d69585aef5c/graphene/relay/connection.py) of the graphene.relay package.

Having iterable in the Connection could be used for tasks like calculating total count for pagination.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,19 +6,19 @@ python:
 - 3.5
 - pypy
 before_install:
-  - |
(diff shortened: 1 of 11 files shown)


syrusakbary:
Sorry for merging so late.

Issue #9, opened 2016-10-10
`convert_column_to_enum` doesn't work at all


philiptzou:
This issue is related with three packages: `graphql-core`, `graphene` and `graphene-sqlalchemy`. Actually there are 2 separated problems related with the Enum type:
1. In `convert_column_to_enum`, the returned value should be an instance of `UnmountedType` but not a subclass of `UnmountedType`.
2. [ `ChoiceType` of `sqlalchemy_utils`](https://github.com/kvesteri/sqlalchemy-utils/blob/0.32.9/sqlalchemy_utils/types/choice.py) coerces the result to `Choice` type or Enum subclass type (depend on which `XxxxTypeImpl` it used). But it's more pain in ass since there's no way to override [`GraphQLEnumType`'s `serialize` method](https://github.com/graphql-python/graphql-core/blob/v0.5.2/graphql/type/definition.py#L474) like `GraphQLScalarType` did.

philiptzou:
3 pull-requests were issued in all 3 repositories to fix this problem. Now my `ChoiceType` finally worked.

richmondwang:
any updates on this issue?

erikwrede:
This should be fixed on master.

PR #10, opened 2016-10-10
Serialization function should be applied on `ChoiceType` output

philiptzou:
This fixed https://github.com/graphql-python/graphene-sqlalchemy/issues/9.

### DIFF
--- a/graphene_sqlalchemy/converter.py
+++ b/graphene_sqlalchemy/converter.py
@@ -11,6 +11,12 @@ from graphene.types.json import JSONString
 from .fields import SQLAlchemyConnectionField
 
 try:
+    from enum import Enum as PyEnum
+except ImportError:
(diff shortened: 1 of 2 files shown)


syrusakbary:
Sorry for addressing this so late.

This PR (and related PRs mentioned in #9) didn't got merged because they were based on making enums serialization "dynamic" (meaning modifiable if indicated by the enum configuration) and that might eventually cause some inconsistency as Enum is a fixed concept and it's serialization and parsing should be uniform across any project.

However I think this issue might be easily solved without the need of modifying it's serialization.

Closing this PR in favor of another PR with a test case that exemplifies this issue (without depending on `GraphQLEnum.serialize`) so we make sure can be addressed without the need of modifying `graphql-core` and `graphene`.

PR #11, opened 2016-10-12
allow abstract subclasses of SQLAlchemyObjectType

sjhewitt:
this allows `SQLAlchemyObjectType` to be subclassed so you can provide common functionality. My use case is to override `resolve_id`, but this could/should be extended to allow the abstract subclass to provide field declarations and perhaps common meta options (interfaces, etc)


syrusakbary:
This is now handled automatically in the upcoming Graphene `2.0`. Closing :)

https://github.com/graphql-python/graphene/pull/500

PR #36, opened 2017-02-24
Adding iterable and length to SQLAlchemyConnectionField

yfilali:
It is currently difficult to resolve a total_count field with the SQLAlchemy graphene library.

The example from [swapi](https://github.com/graphql-python/swapi-graphene/blob/master/starwars/schema.py#L18-L22) has a nice little implementation that does not work here since unlike Django, the SQLAlchemyConnectionField does not set neither length nor iterable on the connection returned.

This PR brings length and iterable to SQLAlchemyConnectionField to match DjangoConnectionField.

### DIFF
--- a/graphene_sqlalchemy/fields.py
+++ b/graphene_sqlalchemy/fields.py
@@ -1,16 +1,14 @@
 from functools import partial
 
-from sqlalchemy.orm.query import Query
-
 from graphene.relay import ConnectionField
 from graphene.relay.connection import PageInfo
 from graphql_relay.connection.arrayconnection import connection_from_list_slice
+from sqlalchemy.orm.query import Query
 
 from .utils import get_query
 
 
 class SQLAlchemyConnectionField(ConnectionField):
-
     @property
     def model(self):
         return self.type._meta.node._meta.model
@@ -28,7 +26,7 @@ class SQLAlchemyConnectionField(ConnectionField):
             _len = iterable.count()
         else:
             _len = len(iterable)
-        return connection_from_list_slice(
+        connection = connection_from_list_slice(
             iterable,
             args,
             slice_start=0,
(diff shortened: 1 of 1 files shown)


sebastiandev:
How would you use the length field to be returned along with the query results?
If I just add length on the query, It fails saying the Connection class doesn't have a length field.

wichert:
@yfilali Is there any example of how to use this? You are storing this on the connection field instance now, but that does not make it visible anywhere. Extending PageInfo with a new field and populating that is still needed somewhere I suspect?

yfilali:
I haven't tested this recently, but the original example should still stand: [swapi](https://github.com/graphql-python/swapi-graphene/blob/master/starwars/schema.py#L18-L22)

The official docs show how to add an extra field to a connection subclass: http://docs.graphene-python.org/en/latest/relay/connection/, in that example, you would just add a resolve_extra method there that returns self.length.

wichert:
The graphene documentation for a custom connection makes you think this should work:

```python
class Article(SQLAlchemyObjectType):
    class Meta:
        model = model.Article
        interfaces = (relay.Node, )


class LengthConnection(graphene.Connection):
    class Meta:
        node = Article

    total_count = graphene.Int()

    def resolve_total_count(self, info):
        return self.length


class Query(graphene.ObjectType):
    articles = SQLAlchemyConnectionField(ArticleConnection)
```

but that fails because ArticleConnection is not a subtype of ArticleConnection.

wichert:
I think I am on to something. DjangoObjectType. __init_subclass_with_meta__ has an extra connection_class parameter, and [uses that as connection type (if provided) instead of dynamically creating a new connection class](https://github.com/graphql-python/graphene-django/blob/master/graphene_django/types.py#L7). There is no equivalent for that in SQLAlchemyObjectType.

PR #60, opened 2017-07-28
Make tests run with Python 2 as well

Cito:
SQLAlchemy expects new-style classes.

So the manually mapped class must inherit from object, even though this is not necessary in Python 3.


Cito:
Sorry, the Travis config was not the problems. Closing this PR to create a better one.

PR #78, opened 2017-09-01
Proper support for the SQLAlchemy Enum type

Cito:
SQLAlchemy Enums were converted to Graphene Strings by graphene_sqlalchemy so far. This is a proposed patch to convert them into Graphene Enums instead. Note that postgresql.ENUM do not need special casing since it is a subtype of the generic SQLAlchemy Enum. Also note that the patch support the two variants of defining an SQLAlchemy Enum, namely by using a sequence and by using a Python Enum.

### DIFF
--- a/graphene_sqlalchemy/converter.py
+++ b/graphene_sqlalchemy/converter.py
@@ -116,7 +116,8 @@ def convert_column_to_datetime(type, column, registry=None):
 @convert_sqlalchemy_type.register(types.Integer)
 def convert_column_to_int_or_id(type, column, registry=None):
     if column.primary_key:
-        return ID(description=get_column_doc(column), required=not (is_column_nullable(column)))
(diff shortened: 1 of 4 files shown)


sboisson:
It seems there is a problem if the same Enum is used in several classes, the code attempt to recreate the Enum with the same name..

Also there should be a way to reference the generated graphene.Enum so it can be used in queries and mutation..

Cito:
@sboisson Thanks for the heads up - I will look into these issues this week.

Cito:
@sboisson I have now addressed these two problems in #98.

It should now be possible to use the same enum, and you can refer to the enum either using the registry or using `sqlalchemy_object_type._meta.fields[column_name].type`. You can see how this works in the `test_should_query_well` unit test.

PR #93, opened 2017-11-21
Allow creation of custom connections. Fixes #65.

### LINKED ISSUE #65: Subclassing Connection in 2.0
Hi, in 2.0dev version I can't see the way to subclass `relay.Connection`. Because SQLAlchemyObjectType doesn't have a Connection attribute, the only way I see is to pass connection via the Meta class. Otherwise the connection is automatically created with default Connection class. But I can't specify the subclassed connection in Meta, because it needs a node attribute and I get circular reference. E. g.

```python
class UserConnection(graphene.relay.Connection):
    class Meta:
        node = User

class User(SQLAlchemyObjectType):
    class Meta:
        model = UserModel
        interfaces = (relay.Node, )
        connection = UserConnection
```

So I don't understand what's the point of connection parameter in SQLAlchemyObjectType Meta class? To make it work, I have changed SQLAlchemyObjectType.\_\_init_subclass_with_meta_\_ and introduced connection_type parameter. Now I can make an abstract subclassed Connection and pass it to Meta:

```python
class SQLAlchemyObjectType(ObjectType):
    @classmethod
    def __init_subclass_with_meta__(cls, model=None, registry=None, skip_registry=False,
                                    only_fields=(), exclude_fields=(), connection=None, connection_type=None,
                                    use_connection=None, interfaces=(), id=None, **options):

        ...

        if use_connection and not connection:
            # We create the connection automatically
            connection = (connection_type or Connection).create_type('{}Connection'.format(cls.__name__), node=cls) 

        ...
```

```python
class UserConnection(graphene.relay.Connection):
    class Meta:
        abstract = True


class User(SQLAlchemyObjectType):
    class Meta:
        model = UserModel
        interfaces = (relay.Node, )
        connection_type = UserConnection
```

Maybe I don't understand something and there is an easier way to make it work?

geertjanvdk:
Previously, it was not possible to create custom connections without
monkey patching graphene.relay.Connection. Custom connections are
useful when all connections need a total count for paging, for example.

We change the SQLAlchemyObjectType.__init_subclass_with_meta method
so that when the connection argument is a class, it is used to create
the new connection type.

A test case has been added to check usage of custom connections. The
`conftest.py` was added to allow pytest using database related fixtures
in other places than `test_query.py`.

### DIFF
--- a/graphene_sqlalchemy/tests/conftest.py
+++ b/graphene_sqlalchemy/tests/conftest.py
@@ -1,0 +1,40 @@
+import pytest
+from sqlalchemy import create_engine
+from sqlalchemy.orm import scoped_session, sessionmaker
+
+from .models import Article, Base, Editor, Reporter
+from ..registry import reset_global_registry
+
+db = create_engine('sqlite:///test_sqlalchemy.sqlite3')
+
+
+@pytest.yield_fixture(scope='function')
+def session():
+    reset_global_registry()
+    connection = db.engine.connect()
+    transaction = connection.begin()
+    Base.metadata.create_all(connection)
+
+    # options = dict(bind=connection, binds={})
+    session_factory = sessionmaker(bind=connection)
+    session = scoped_session(session_factory)
+
+    yield session
+
+    # Finalize test here
+    transaction.rollback()
+    connection.close()
+    session.remove()
+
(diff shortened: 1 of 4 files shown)


geertjanvdk:
I'm closing this request and reopen with rebased branch.

PR #101, opened 2017-12-16
Added sort support

### LINKED ISSUE #40: Order by option is missing
Great job,
Order by option is missing.

CaselIT:
Add support for sorting the return values of a `SQLAlchemyConnectionField`

The `sort` argument of `SQLAlchemyConnectionField` supports sorting over a single over multiple field.

In the utils file I've added
-  the function `sort_enum_for_model` to create a graphene Enum with all be possible field and order combination for a particular sqlalchemy model as suggested by @Cito 
- the function `sort_argument_for_model` to automacally create the enum and return the attribute to use in the connection field

The test and documentation are still missing, because first I wanted to get some feedback about the implementation.

Regarding the value of the sort enumerator at first I tried to set the values of the enum to the ascenting or descenting operators of sqlalchemy simplify the logic of the get_query but that results in an error when graphql builds the ast graph, so I reverted to use a list column_name, operator that is less than optimal since every time it needs to query the model to create the operators.

Thanks to @Cito to the input in #40 
Closes #40

### DIFF
--- a/docs/examples.rst
+++ b/docs/examples.rst
@@ -3,7 +3,7 @@ Schema Examples
 
 
 Search all Models with Union
------------------
+----------------------------
 
 .. code:: python
 
@@ -11,12 +11,22 @@ Search all Models with Union
         class Meta:
             model = BookModel
             interfaces = (relay.Node,)
-    
-    
+
+
+    class BookConnection(relay.Connection):
+        class Meta:
+            node = Book
+
+
     class Author(SQLAlchemyObjectType):
         class Meta:
             model = AuthorModel
             interfaces = (relay.Node,)
+
+
+    class AuthorConnection(relay.Connection):
+        class Meta:
+            node = Author
 
 
     class SearchResult(graphene.Union):
@@ -29,8 +39,8 @@ Search all Models with Union
         search = graphene.List(SearchResult, q=graphene.String())  # List field for search results
         
         # Normal Fields
-        all_books = SQLAlchemyConnectionField(Book)
-        all_authors = SQLAlchemyConnectionField(Author)
+        all_books = SQLAlchemyConnectionField(BookConnection)
+        all_authors = SQLAlchemyConnectionField(AuthorConnection)
 
         def resolve_search(self, info, **args):
             q = args.get("q")  # Search query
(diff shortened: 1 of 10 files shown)


CaselIT:
> Regarding the value of the sort enumerator at first I tried to set the values of the enum to the ascenting or descenting operators of sqlalchemy simplify the logic of the get_query but that results in an error when graphql builds the ast graph, so I reverted to use a list column_name, operator that is less than optimal since every time it needs to query the model to create the operators.
I've changed the implementation to use a subclass of string instead of the list, so that the ast graph can be created successfully and there is no need to query the model every time.

Cito:
Nice, thank you.

Maybe, as I've already commented in #40, the `sort` argument should be added by default to the `SQLAlchemyConnectionField`, so that you don't need to specify it manually. That would be similar to how a Graphene `ConnectionField` comes with `first`, `last` arguments by default, which also don't need to added manually.

CaselIT:
Good idea. I'll add it.
Do you think it should have a default value when added by default?

Cito:
For an `SQLAlchemyConnectionField` it probably makes sense to use the `id` as default value for the sort argument so that you always get a reproducable order.

CaselIT:
`SQLAlchemyConnectionField` now adds the sort attribute by default, with a default sort over the primary key(s). To disable this field pass to the constructor `sort=None`

Kazanz:
This going to get merged?

CaselIT:
I home so.
@syrusakbary could you review this PR?
Thanks

CaselIT:
@syrusakbary see https://github.com/graphql-python/graphene-sqlalchemy/pull/104#issuecomment-361208698

YehonatanVonage:
Hi, 
Any updates about this PR?

CaselIT:
On my part I wanted a feedback from @syrusakbary about the implementation before writing the tests.

millicentachieng:
@syrusakbary We are waiting on you to review the PR

syrusakbary:
Sorry for the time taken on reviewing this PR, and thanks for the ping.
It looks good, once it have tests related to the new functionality I'll merge :)

CaselIT:
Thanks for the review @syrusakbary 
I'll work on adding the tests and ping you once they are done

CaselIT:
@syrusakbary I think it's ready for a review
I've added the tests
Updated to support graphene v2
Added a documentation example and updated some other documentation to be compatible with v2

Let me know what you think

syrusakbary:
It looks great! You did a great job 👏 

Merging

CaselIT:
Thank you for merging it!

PR #105, opened 2017-12-20
Update converter.py again to include INET and CIDR types.

scotmatson:
Type coverage for this is poor and needs to address a wider array of types provided by SQLAlchemy.


scotmatson:
This PR replaces my previous PR of the same issue, #69 which began failing changes in the code base following the request.

PR #147, opened 2018-07-24
MySQL Enum conversion issue

sebastiandev:
When using reflection to build the models (instead of declarative) the Enum conversion fails because the ENUM column is treated as string so doing type.name returns None.

```sql
CREATE TABLE `foos` (
  `id` int(11) unsigned NOT NULL AUTO_INCREMENT,
  `status` ENUM('open', 'closed') DEFAULT 'open',
  PRIMARY KEY (`id`)
) ENGINE=InnoDB DEFAULT CHARSET=utf8;
```

```python
class Foo(Base):

    __table__ = Table('foos', metadata, autoload=True)

```

For some reason these kind of enums are returned as Strings so in the type definition we have to specify the field as string

```python
class FooType(SQLAlchemyObjectType):
    class Meta:
        model = Foo
        interfaces = (relay.Node, )
  
    status = String()
```

PR #154, opened 2018-08-03
Fix creation of graphene.Enum from enum.Enum

### LINKED ISSUE #152: Serializing native Python enums does not work
Currently using SQL enums works well, unless you use a `enum.Enum` as base for the enum. For example using this model:

```python
class Hairkind(enum.Enum):
    LONG = 'long'
    SHORT = 'short'

class Pet(Base):
    __tablename__ = 'pets'
    id = Column(Integer(), primary_key=True)
    hair_kind = Column(Enum(Hairkind, name='hair_kind'), nullable=False)
```

will fail badly if you try to query the hairKind field:

```
  File "lib/python3.7/site-packages/graphql/execution/executor.py", line 622, in complete_leaf_value
    path=path,
graphql.error.base.GraphQLError: Expected a value of type "hair_kind" but received: Hairkind.LONG
```

wichert:
This fixes #152, and adds a test to make sure it keeps working :)

### DIFF
--- a/graphene_sqlalchemy/converter.py
+++ b/graphene_sqlalchemy/converter.py
@@ -146,12 +146,14 @@ def convert_column_to_float(type, column, registry=None):
 
 @convert_sqlalchemy_type.register(types.Enum)
 def convert_enum_to_enum(type, column, registry=None):
-    try:
-        items = type.enum_class.__members__.items()
-    except AttributeError:
+    enum_class = getattr(type, 'enum_class', None)
+    if enum_class:  # Check if an enum.Enum type is used
+        graphene_type = Enum.from_enum(enum_class)
+    else:  # Nope, just a list of string options
         items = zip(type.enums, type.enums)
+        graphene_type = Enum(type.name, items)
     return Field(
-        Enum(type.name, items),
+        graphene_type,
         description=get_column_doc(column),
         required=not (is_column_nullable(column)),
     )
--- a/graphene_sqlalchemy/tests/models.py
+++ b/graphene_sqlalchemy/tests/models.py
@@ -6,6 +6,12 @@ from sqlalchemy import Column, Date, Enum, ForeignKey, Integer, String, Table
 from sqlalchemy.ext.declarative import declarative_base
 from sqlalchemy.orm import mapper, relationship
 
+
+class Hairkind(enum.Enum):
+    LONG = 'long'
+    SHORT = 'short'
+
+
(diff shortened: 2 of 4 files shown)


wichert:
There is a bit of an issue here, which was causing the tests to fail: using Enum.from_enum() makes the conversion to and from PyEnum works correctly, but it looses the name from the SQLAlchemy enum type and instead uses the name from the PyEnum type. I would love to argue that that is the Right Thing To Do ™️, but I can see it being a surprise to people. Fixing it probably means extending graphene to support a name parameter to Enum.from_enum().

@syrusakbary, can you weigh in on this?

stan-sack:
This PR does not fix my `graphene.InputObjectType` which was defined as:

```
class PetInput(graphene.InputObjectType):
    ...
   hairKind = graphene.Enum.from_enum(models.HairKind)
    ...
```

The InputObjectType above works in v2.0.0.

wichert:
@stan-sack Is that another regression in v2.1.0? It looks like there are no tests for enum inputs, so we'll need to add those I guess.

sebastiandev:
We had an issue today and this worked, it only needs an extra change from my [PR](https://github.com/graphql-python/graphene-sqlalchemy/pull/147)

```python
@convert_sqlalchemy_type.register(types.Enum)
def convert_enum_to_enum(type, column, registry=None):
    enum_class = getattr(type, 'enum_class', None)

    if enum_class:  # Check if an enum.Enum type is used
        graphene_type = Enum.from_enum(enum_class)

    else:  # Nope, just a list of string options
        items = zip(type.enums, type.enums)
        graphene_type = Enum(type.name or "{}_{}".format(column.table, column.name), items)

    return Field(
        graphene_type,
        description=get_column_doc(column),
        required=not (is_column_nullable(column)),
    )

```

kousun12:
@wichert any chance we can get this in soon?

wichert:
@kousun12 That is up to @syrusakbary - this is how project. Personally I ended up reevaluating our use of GraphQL and graphene and moved back to REST.

syrusakbary:
My fault. These last months had been frenetic personally for me (new company creation, new immigrant visa to be able to stay in the US, ...) and it had occasioned some downsides/side-effects to Graphene.
Sorry about that, things will start getting into shape soon.

PS: To everyone using the project. If you are using it commercially, consider donations as a way to support it, so it can become a sustainable project: https://www.patreon.com/syrusakbary

PR #162, opened 2018-08-30
fixes #153: naming conflict when using Connection suffix

### LINKED ISSUE #153: flask_sqlalchemy example : AssertionError: Found different types with the same name in the schema: EmployeeConnection, EmployeeConnection.
Hi,

I followed the instruction to use the flask_sqlalchemy with a Python 2.7 version, when I execute python app.py, I've got the following error : 

`
(py27) λ python app.py                                                                                                                      
  File "app.py", line 7, in <module>                                                                                                        
    from schema import schema                                                                                                               
  File "D:\Thomas\Dev\graphene-sqlalchemy\examples\flask_sqlalchemy\schema.py", line 60, in <module>                                        
    schema = graphene.Schema(query=Query, types=[Department, Employee, Role])         

....
                            
  File "D:\Programmes\Anaconda2\envs\py27\lib\site-packages\graphene\types\typemap.py", line 99, in graphene_reducer                        
    ).format(_type.graphene_type, type)                                                                                                     
AssertionError: Found different types with the same name in the schema: EmployeeConnection, EmployeeConnection.                             
`

What am I doing wrong ? 

Thank you

tomahim:
I propose to fix the #153 issue by renaming all connection classes with the suffix `Connection` as it seems to be use by the `graphene_reducer`.

The suffix `Conn` may not be great, I'm open to other suggestions 😄

### DIFF
--- a/examples/flask_sqlalchemy/schema.py
+++ b/examples/flask_sqlalchemy/schema.py
@@ -12,7 +12,7 @@ class Department(SQLAlchemyObjectType):
         interfaces = (relay.Node, )
 
 
-class DepartmentConnection(relay.Connection):
+class DepartmentConn(relay.Connection):
     class Meta:
         node = Department
 
@@ -23,7 +23,7 @@ class Employee(SQLAlchemyObjectType):
         interfaces = (relay.Node, )
 
 
-class EmployeeConnection(relay.Connection):
+class EmployeeConn(relay.Connection):
     class Meta:
         node = Employee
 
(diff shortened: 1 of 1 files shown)


swuecho:
thanks! saved my day.

mfrlin:
This does not solve underlying issue.

jnak:
The issue has not been fixed in the sense that it does not need to be fixed. The way to workaround this is to avoid defining Connection for types. If you need to refer to those connections, you can access it with `YourType._meta.connection`. Eventually, we should expose a way to get it without accessing a private attribute.

Closing this since this is not correct approach.

jnak:
See https://github.com/graphql-python/graphene-sqlalchemy/issues/153#issuecomment-528373570 for more context

PR #165, opened 2018-09-27
Fix for enum type duplication in graphene schema

hoffrocket:
The converter code was created a new type for each Enum column that it
encountered. The problem is where the same Enum type is used for
multiple columns. In that case graphene will notice two schema types
with the same underlying name and throw an Assertion error like:

AssertionError: Found different types with the same name in the schema: pet_kind, pet_kind.

This fix creates a new registry dict for storing a type.name to Enum
mapping that allows for reuse.

### DIFF
--- a/graphene_sqlalchemy/converter.py
+++ b/graphene_sqlalchemy/converter.py
@@ -146,12 +146,20 @@ def convert_column_to_float(type, column, registry=None):
 
 @convert_sqlalchemy_type.register(types.Enum)
 def convert_enum_to_enum(type, column, registry=None):
-    try:
-        items = type.enum_class.__members__.items()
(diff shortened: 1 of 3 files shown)


hoffrocket:
I notice that there are some other outstanding PRs that address this issue in addition to some other problems with enums. Any reason those other changes haven't been merged? Would love to find something to address this. Currently have a workaround in place that registers a this solution using @convert_sqlalchemy_type.register(sqlalchemy.types.Enum)

Nabellaleen:
Hi @hoffrocket !

graphene-sqlalchemy maintainer team is now alive ! Could you rebase and resolve conflicts ? :)

PR #166, opened 2018-10-03
Added a new example using Falcon and graphene-sqlalchemy

somada141:
Some time ago I put together an example utilising [Falcon](https://falconframework.org/) and graphene-sqlalchemy. It can be found under [https://github.com/somada141/demo-graphql-sqlalchemy-falcon](https://github.com/somada141/demo-graphql-sqlalchemy-falcon).

The example demonstrates setting up the SQLAlchemy ORM schema, representing it in GraphQL through graphene-sqlalchemy, and performing queries and mutations.

It also demonstrates how to use the GraphQL AST to limit the SQLAlchemy queries to only the fields requested in the GraphQL query.


somada141:
Just realised the build fails cause its trying to run the tests out of my example, will fix

FabriceSalvaire:
Thanks to share ! It's an interesting use case.

PR #168, opened 2018-10-03
Added an `Aggregation` section to the `tips` documentation page.

somada141:
Added an `Aggregation` section to `tips.rst` to show how to retrieve the `session` out of `info` and perform an aggregation query wrapping the results in a custom type and exposing it to the schema.

### DIFF
--- a/docs/tips.rst
+++ b/docs/tips.rst
@@ -85,3 +85,186 @@ some of the allowed queries are
         }
     }
 
+
+Aggregating
+-----------
+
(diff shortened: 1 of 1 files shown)


Nabellaleen:
@somada141 Nice example, but is it really related to graphene ?

The "tips" is "how to retrieve easily the sqlalchemy session from graphene context", but all the "what you can do with the query" is more sqlalchemy stuff than graphene stuff, no ?
So maybe most of the code you wrote should go as example in the appropriated directory instead of in the tips ?

(we can merge that this way and create a fully functional aggragation example later)

PR #169, opened 2018-10-12
Don't suppress SQLAlchemy errors when mapping classes

### LINKED ISSUE #121: SQLAlchemyObjectType hides original SQLAlchemy exception
Hi all!

Here's a minimal code that explains the issue.

```py
import graphene
from graphene_sqlalchemy import SQLAlchemyObjectType
from sqlalchemy import Column, Integer
from sqlalchemy.ext.declarative import declarative_base
from sqlalchemy.orm import relationship

Base = declarative_base()


class A(Base):
    __tablename__ = 'a'
    id = Column(Integer, primary_key=True)


class B(Base):
    __tablename__ = 'b'
    id = Column(Integer, primary_key=True)

    a = relationship(A)  # This is the breaking line


class BSchema(SQLAlchemyObjectType):
    class Meta:
        model = B


schema = graphene.Schema(query=BSchema)
```

When I try to run this file I only get this error:
```
  File "issue.py", line 22, in <module>
    class BSchema(SQLAlchemyObjectType):
  File "./.venv/lib/python3.6/site-packages/graphene/utils/subclass_with_meta.py", line 40, in __init_subclass__
    super_class.__init_subclass_with_meta__(**options)
  File "./.venv/src/graphene-sqlalchemy/graphene_sqlalchemy/types.py", line 97, in __init_subclass_with_meta__
    ).format(cls.__name__, model)
AssertionError: You need to pass a valid SQLAlchemy Model in BSchema.Meta, received "<class '__main__.B'>".
```

But I'm passing a valid SQLAlchemy Model in BSchema.Meta.

The problem is actually another. 
SQLAlchemy is raising this exception: 
```{NoForeignKeysError}Could not determine join condition between parent/child tables on relationship B.a - there are no foreign keys linking these tables.  Ensure that referencing columns are associated with a ForeignKey or ForeignKeyConstraint, or specify a 'primaryjoin' expression.```

But this exception is lost here https://github.com/graphql-python/graphene-sqlalchemy/blob/a2fe9263d73b3baccd63b192b983cda5abfb88f0/graphene_sqlalchemy/utils.py#L21-L27 (called from https://github.com/graphql-python/graphene-sqlalchemy/blob/a2fe9263d73b3baccd63b192b983cda5abfb88f0/graphene_sqlalchemy/types.py#L94)
making the debug process for a simple mistake a pain.

### PR DESCRIPTION
**Pull request by connorbrinton**
Currently, many SQLAlchemy model definition errors are hidden the way we check if valid models have been provided. This PR allows SQLAlchemy model definition errors to propagate at model check time, and adds a test to confirm that errors are correctly propagated.

Fixes #121

### DIFF
--- a/graphene_sqlalchemy/utils.py
+++ b/graphene_sqlalchemy/utils.py
@@ -27,7 +27,13 @@ def get_query(model, context):
 def is_mapped_class(cls):
     try:
         class_mapper(cls)
-    except (ArgumentError, UnmappedClassError):
+    except ArgumentError as error:
+        # Only handle ArgumentErrors for non-class objects
+        if "Class object expected" in str(error):
+            return False
+        raise
+    except UnmappedClassError:
+        # Unmapped classes return false
         return False
     else:
         return True
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,7 +1,7 @@
 default_language_version:
     python: python3.7
 repos:
--   repo: git://github.com/pre-commit/pre-commit-hooks
+-   repo: https://github.com/pre-commit/pre-commit-hooks
     rev: c8bad492e1b1d65d9126dba3fe3bd49a5a52b9d6  # v2.1.0
     hooks:
     -   id: check-merge-conflict
@@ -11,15 +11,15 @@ repos:
         exclude: ^docs/.*$
     -   id: trailing-whitespace
         exclude: README.md
--   repo: git://github.com/PyCQA/flake8
+-   repo: https://github.com/PyCQA/flake8
     rev: 88caf5ac484f5c09aedc02167c59c66ff0af0068  # 3.7.7
     hooks:
     -   id: flake8
--   repo: git://github.com/asottile/seed-isort-config
+-   repo: https://github.com/asottile/seed-isort-config
     rev: v1.7.0
     hooks:
     -   id: seed-isort-config
--   repo: git://github.com/pre-commit/mirrors-isort
+-   repo: https://github.com/pre-commit/mirrors-isort
     rev: v4.3.4
     hooks:
     -   id: isort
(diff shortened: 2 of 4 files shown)

### CODE REVIEW

**Comment by indigoviolet** [comment]
is this blocked?

**Comment by connorbrinton** [comment]
Nope, everything should be good to go! I just rebased on master to make sure it's up to date with the latest changes as well.

**Comment by lanshark** [comment]
I implemented these changes in a local virtualenv to test an error I was getting.  I can confirm that these changes DO indeed fix the problem.

**Comment by alefhsousa** [comment]
Hi guys, I got this error today. We have any problem to increase these changes on master?

File: graphene_sqlalchemy/utils.py [hunk 1]
```diff
@@ -25,8 +25,17 @@ def get_query(model, context):
 def is_mapped_class(cls):
     try:
         class_mapper(cls)
-    except (ArgumentError, UnmappedClassError):
-        return False
+    except (ArgumentError, UnmappedClassError) as error:
```
Nabellaleen: I think it could be more pythonic written this way : 
```
    except ArgumentError as error:
        # Only handle ArgumentErrors for non-class objects
        if "Class object expected" in str(error):
            return False
        raise
    except UnmappedClassError:
        return False
```

Nabellaleen (dismissed):

graphene_sqlalchemy/utils.py
heller: I appreciate your attention to detail, but is this nitpick worth keeping all of us users stuck on incorrect error messages for months?

jnak: @heller It looks like there are merge conflicts so I think it's fine to address those nitpicks when fixing them. If you want to fix it and send a PR, I'm happy to merge it asap. Cheers.

erikwrede (approved):

PR #170, opened 2018-10-12
Remove unused import.

alexBaizeau:
Remove an unused import from the example


somada141:
@alexBaizeau tests seem to fail on Py3.4 but it doesn't seem related to the code.

Nabellaleen:
Build should be fixed on 3.4 soon by #177

PR #178, opened 2018-11-26
Add a meta option to allow field names renaming

Nabellaleen:
Example is given in the README.md and in tests

It allows to implement : 

```
class User(SQLAlchemyObjectType):
    class Meta:
        model = UserModel
        aliased_fields = {'name': 'first_name'}
```
To be able to have a field named `first_name` in your graphql schema instead of being stucked with UserModel field name `name`

### DIFF
--- a/graphene_sqlalchemy/tests/test_types.py
+++ b/graphene_sqlalchemy/tests/test_types.py
@@ -183,7 +183,7 @@ def test_promise_connection_resolver():
     assert result is not None
 
 
-class HumanWithFieldAliased(SQLAlchemyObjectType):
+class HumanWithFieldRenamed(SQLAlchemyObjectType):
 
     publication_timestamp = Int()
 
@@ -191,7 +191,7 @@ class HumanWithFieldAliased(SQLAlchemyObjectType):
         model = Article
         registry = registry
         interfaces = (Node,)
-        aliased_fields = {
+        rename_fields = {
             "id": "article_id",
             "headline": "title",
             "pub_date": "publication_timestamp",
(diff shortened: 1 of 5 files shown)


Nabellaleen:
Existing tests were not covering lines I've edited and I admit I don't know how to test its (I don't know what are composite fields and hybrid fields)

If someone want to help me on that, it's with pleasure ;)

Cito:
Nice. This might be useful when you can't change the SQLAlchemy model or are using reflection.

But maybe the name "aliased_fields" is a bit misleading, because an alias is usually an alternative, *additional* name that exists parallel to the original name, but what you're doing is *renaming* the fields. Also, the other attributes use the imperative form "skip_", "use_", "exclude_", instead of "skipped_", "used_", "exclude_". So to be more clear and consistent I suggest the name "rename_fields" instead.

graphene_sqlalchemy/tests/test_types.py
```diff
@@ -181,3 +181,36 @@ class Meta:
...
+            "pub_date": "publication_timestamp",
+            "reporter_id": "journalist_id",
+            "reporter": "journalist",
+        }
+
+
+def test_objecttype_with_aliased_fields():
+    assert issubclass(HumanWithFieldAliased, ObjectType)
+    assert HumanWithFieldAliased._meta.model == Article
+    assert list(HumanWithFieldAliased._meta.fields.keys()) == [
```
Cito: You don't need the `.keys()` here because a dict is iterable itself. Normally you would need to sort the keys to get a robust test, but in this case we have an ordered dict so this is ok.

graphene_sqlalchemy/tests/test_types.py
```diff
@@ -181,3 +181,36 @@ class Meta:
...
+    publication_timestamp = Int()
+
+    class Meta:
+        model = Article
+        registry = registry
+        interfaces = (Node,)
+        aliased_fields = {
+            "id": "article_id",
+            "headline": "title",
+            "pub_date": "publication_timestamp",
```
Cito: Maybe don't rename *all* keys but leave one original name to test if that works, too.

Cito: Maybe also add a similar test to `test_reflected.py` because that is the case where this is most useful in practice.

jnak:
Closing this since it is now possible to rename fields via [`ORMField.model_attr`](https://github.com/graphql-python/graphene-sqlalchemy/blob/master/graphene_sqlalchemy/types.py#L28-L74). This will be more thoroughly documented soon. For now, you can also refer to the some of the [tests](https://github.com/graphql-python/graphene-sqlalchemy/blob/631513fe42cb0b0613349f17118faabf879107ac/graphene_sqlalchemy/tests/test_types.py#L309).

PR #183, opened 2019-01-22
Use a conditional dependency for singledispatch

### LINKED ISSUE #182: singledispatch is not necessary for python_version >= Python3.4
I am using Python 3.7 and the attempt to install singledispatch is causing my build to break. I believe the way to resolve this issue is by using [conditional dependencies](https://hynek.me/articles/conditional-python-dependencies/) as outlined in [PEP 508](https://www.python.org/dev/peps/pep-0508/). I will follow up with a PR shortly.

smfoote:
I am not very familiar with conditional dependencies in Python, so please let me know if that is not the right way to handle this.

Fixes #182

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -40,7 +40,7 @@ setup(
         "six>=1.10.0",
         "graphene>=2.1.3",
         "SQLAlchemy",
-        "singledispatch>=3.4.0.3",
+        "singledispatch>=3.4.0.3 ; python_version<'3.4'",
     ],
     tests_require=["pytest>=2.7.2", "mock", "sqlalchemy_utils"],
 )


Nabellaleen:
Please check my question on #182 ;)

If your answer shows the conditional dependencie is required, you should also write a conditional import in `graphene_sqlalchemy/converter.py`, as the `singledispatch` in the standard library should be import with : `from functools import singledispatch`

```
try:
    from functools import singledispatch
except ImportError:
    from singledispatch import singledispatch
```

dsully:
Would love to see this change updated, pushed and a release made. Our internal build system excludes "singledispatch" from our dependency tree as we are Python 3-only.

Thanks!

JBrVJxsc:
having the same problem here. cannot believe this is not fixed after 2 years...

erikwrede:
This is resolved in 3.0-Beta. The singledispatch dependency was removed since support for python < 3.6 was dropped.
If you require additional changes to be made, please open a new PR.

PR #177, opened 2018-11-26
Use enum34 backport when Enum is not available

Nabellaleen:
Tests were broken in Python < 3.4, this PR fix its

PR #187, opened 2019-03-25
Pass `relationship` and `registry` objects to `connection_field_factory`

jnak:
Hello,

The `createConnectionField` currently only expects the `type` of the connection. I found this to be limiting since there are valid cases where connection factories depend on other attributes of the `relationship` object.

For example, in my case I need the connection_factory to inspect the `relationship` object to extract the `JoinCondition` object from `SqlAlchemy`. This allows us to auto-generate performant sql-based connections for SqlAlchamy `relationships` that only fetch the relevant page of data from the database <sup>1</sup>. 

Given `convert_sqlalchemy_relationship` takes both a `relationship` and a `registry` objects, I think it makes sense to have the factory take the same parameters. That said, I think it would be more future proof to also pass the `model` object passed in `construct_fields`.

Thank you!
J

_ 
<sup>1</sup> I will send a PR in the next few weeks to see if there is any interest to share this here.


graphene_sqlalchemy/fields.py
```diff
@@ -94,18 +98,46 @@ def __init__(self, type, *args, **kwargs):
...
 
 
-__connectionFactory = UnsortedSQLAlchemyConnectionField
+def default_connection_field_factory(relationship, registry):
+    model = relationship.mapper.entity
+    _type = registry.get_type_for_model(model)
+    return UnsortedSQLAlchemyConnectionField(_type._meta.connection)
+
+
+__connection_field_factory =  default_connection_field_factory
```
jnak: The current pattern to register a factory is to use a global module variable. I followed it in this PR but I think we should move away from it because:
- unless every module that defines a subclass of SQLAlchemyObjectType explicitly calls `register_connection_field_factory`, you may not get the expected connection when you import a `SQLAlchemyObjectType` subclass. 
- it is not possible to use different factories within the same schema.

Instead I would prefer to explicitly define the factory on the `Meta` class of `SQLAlchemyObjectType`:
```python
class MyType(SQLAlchemyObjectType):
     class Meta:
         connection_field_factory = ...
         ...
```

This behavior could easily be shared within an app by putting this in an abstract meta class. Let me know if you agree with this change and I'll update this PR.

graphene_sqlalchemy/tests/test_connectionfactory.py
```diff
@@ -1,44 +1,69 @@
...
-            class Meta:
-                node = table
-
-        return LXConnectionField(
-            LXConnection,
-            filter=table.filter(),
-            order_by=graphene.List(of_type=table.order_by),
-        )
-
-    registerConnectionFieldFactory(createLXConnectionField)
```
jnak: FYI those tests were not testing anything since not schema was defined

graphene_sqlalchemy/fields.py
```diff
@@ -94,18 +98,46 @@ def __init__(self, type, *args, **kwargs):
...
+
+
+# TODO Remove in next major version
 
 def registerConnectionFieldFactory(factoryMethod):
-    global __connectionFactory
-    __connectionFactory = factoryMethod
+    log.warn(
+        'registerConnectionFieldFactory is deprecated and will be removed in the next '
+        'major version. Use register_connection_field_factory instead.'
```
jnak: I created new functions using camelcase because it seems to be the pattern followed in this project.

setup.py
```diff
@@ -42,5 +42,5 @@
         "SQLAlchemy",
         "singledispatch>=3.4.0.3",
     ],
-    tests_require=["pytest>=2.7.2", "mock", "sqlalchemy_utils"],
+    tests_require=["pytest>=2.7.2", "mock", "sqlalchemy_utils", "enum34"],
```
jnak: FYI this is needed to run test with python 2.7. I was not able to run the test with python 3 with and without this.

jnak: I ended up removing it because of https://github.com/graphql-python/graphene-sqlalchemy/pull/177 already fixes it.

graphene_sqlalchemy/fields.py
Nabellaleen: Clearly it seems more "standard" to implement it as meta parameter

But think also to retro-compatibility if you change the behavior (with deprecated warnings as you wrote for `register_connection_field_factory`)
Maybe you could do it in another PR, because this is a different subject

graphene_sqlalchemy/fields.py
```diff
@@ -94,18 +98,46 @@ def __init__(self, type, *args, **kwargs):
...
+    _type = registry.get_type_for_model(model)
+    return UnsortedSQLAlchemyConnectionField(_type._meta.connection)
+
+
+__connection_field_factory =  default_connection_field_factory
 
+def create_connection_field(relationship, registry):
+    return __connection_field_factory(relationship, registry)
 
-def createConnectionField(_type):
```
Nabellaleen: Maybe you should also keep `createConnectionField` with a deprecated warning ?

graphene_sqlalchemy/tests/test_connectionfactory.py
Nabellaleen: So you tests are covering the same area as intended for these tests ?

graphene_sqlalchemy/fields.py
jnak: My worry is that if we merge this as is now we'll have to support 2 deprecated methods (`register_connection_field_factory` and `registerConnectionFieldFactory) with different interfaces. In my opinion it would easier / cleaner to fix everything at once if you also think that 'meta' is the way to go. What do you think?

graphene_sqlalchemy/tests/test_connectionfactory.py
jnak: Yeah my tests are testing that `register_connection_field_factory` and `unregister_connection_field_factory` actually work.

graphene_sqlalchemy/fields.py
jnak: Fair point. This function is supposed to be private to the library but we can't guarantee that.

graphene_sqlalchemy/fields.py
Nabellaleen: Clearly, I think the "meta" approach is the right way.

My point is just : if you implement this with the "meta" way, do you still keep a "retro-compatibility" on the old `registerConnectionFieldFactory` ? I think it's important

jnak:
@Nabellaleen I made the changes we discussed. Would you mind taking a look at it? Thanks

graphene_sqlalchemy/converter.py
```diff
@@ -23,7 +21,7 @@ def is_column_nullable(column):
     return bool(getattr(column, "nullable", True))
 
 
-def convert_sqlalchemy_relationship(relationship, registry):
+def convert_sqlalchemy_relationship(relationship, registry, connection_field_factory):
```
Nabellaleen: Maybe you should add a default value, instead of `default_connection_field_factory` everywhere ?

graphene_sqlalchemy/types.py
```diff
@@ -14,11 +14,12 @@
                         convert_sqlalchemy_composite,
                         convert_sqlalchemy_hybrid_method,
                         convert_sqlalchemy_relationship)
+from .fields import default_connection_field_factory
 from .registry import Registry, get_global_registry
 from .utils import get_query, is_mapped_class, is_mapped_instance
 
 
-def construct_fields(model, registry, only_fields, exclude_fields):
+def construct_fields(model, registry, only_fields, exclude_fields, connection_field_factory):
```
Nabellaleen: same thing about default here ?

graphene_sqlalchemy/converter.py
jnak: The default it set on the main `SQLAlchemyObjectType` class. Since this function is only called by this class (via the  `construct_fields` function), I think it makes sense to keep the default logic there in one place. I don't think that `convert_sqlalchemy_relationship` nor `construct_fields` should know about the default factory.

graphene_sqlalchemy/types.py
jnak: I addressed this in your other comment

PR #197, opened 2019-04-01
Fix #196 - handle ChoiceType initialized with Enum

### LINKED ISSUE #196: Enum ChoiceType column results in error
Howdy folks!

The following code produces `TypeError: 'MyEnum' object is not iterable`, but should work. Note that this is also involves package `sqlalchemy_utils`

```py
from enum import Enum

class MyEnum(Enum):
   foo = 1
   bar = 2

class MyModel(Base):
    id = Column(sa.Integer, primary_key=True)
    name = Column(sa.Unicode(255))
    type = Column(ChoiceType(MyEnum, impl=Integer()))

class MyObject(SQLAlchemyObjectType):
    class Meta:
        model = MyModel
```

thejcannon:


graphene_sqlalchemy/converter.py
```diff
@@ -162,7 +162,12 @@ def convert_enum_to_enum(type, column, registry=None):
 @convert_sqlalchemy_type.register(ChoiceType)
 def convert_column_to_enum(type, column, registry=None):
     name = "{}_{}".format(column.table.name, column.name).upper()
-    return Enum(name, type.choices, description=get_column_doc(column))
+    enum_members = getattr(type.choices, "__members__", None)
+    if enum_members:  # Check if an enum.Enum is used
+        items = [(item.name, item.value) for item in enum_members.values()]
```
thejcannon: We could also return `Enum.from_enum(type.choices)`.
I don't know enough about what the expected name/description should be to know which way is more correct.

thejcannon:
Closing. I'm going to post a new PR that'll fix this and #199 by modeling the code after `convert_enum_to_enum`

thejcannon:
I reopened as #199 isn't seemingly an issue with this library.

thejcannon:
Closing again in lieu of new PR to fix enums in general

squarewave24:
hello, 
i came across this issue on graphene-sqlalchemy v2.2.2
```
TypeError: cannot unpack non-iterable ReviewStatusEnum object
```
on
```
status = Column(ChoiceType(ReviewStatusEnum, impl=Unicode()), nullable=False)
```

is there a trick to make this work?

jnak:
I'm not familiar with the issue but if you need a quick workaround, you should be able to define your own graphene field, enum and resolver.

squarewave24:
actually it was resolved in latest pre-release version 2.3.* so i am testing that out now.

PR #198, opened 2019-04-01
Update version to 2.1.1

### LINKED ISSUE #184: Create a new release which includes #154
The last release was quite some time ago. Shortly after the release #154 was implemented which contains a fix for Enum creation. Would it be possible to create a new release including the changes since the last relase?

Nabellaleen:
Fix #184 
- [x] Change version in __init__.py
- [x] Fix Pypi deploy on Travis


jnak:
Looking at what Syrus usually does in other repos using Tox, the build is expected to fail when releasing on Pypi. We just need one environment to work.

https://travis-ci.org/graphql-python/graphene/builds/406016097

PR #200, opened 2019-04-02
Added description to relationships

sedough:
Generates a description for a SQLAlchemy relationship field using the "doc" attribute.


graphene_sqlalchemy/converter.py
```diff
@@ -35,7 +35,7 @@ def dynamic_type():
         elif direction in (interfaces.ONETOMANY, interfaces.MANYTOMANY):
             if _type._meta.connection:
                 return createConnectionField(_type._meta.connection)
```
jnak: It seems like createConnectionField could benefit from this as well

MrEbbinghaus:
@jnak Any progress on this one?

PR #207, opened 2019-04-15
Release 2.1.2 

jnak:
Release https://github.com/graphql-python/graphene-sqlalchemy/pull/187.

Also add documentation for the release process.

Btw we should decide as a group if we want to follow Semver or if we want to somehow track `graphene` version. I'm in favor of the former but it seems we are currently following the latter.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,0 +1,31 @@
+## Local Development
+
+Set up our development dependencies:
+
+```sh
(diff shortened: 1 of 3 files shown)


jnak:
Here is a draft of the release notes: https://github.com/graphql-python/graphene-sqlalchemy/releases/edit/untagged-9f5da202a45353833ab1

Issue #212, opened 2019-05-09
Generate Mutation input from Model 


Rafik-Belkadi:
Is there a way to have something like this ? 

```python
class AddUserInput(SQLAlchemyInputObjectType):
    class Meta:
        model = UserModel

```

jnak:
Not at the moment. If you're interested in contributing to this, feel free to start a design conversation outlining your approach. Cheers.

thejcannon:
FWIW I've started basing my mutations off methods in my model, with the hopes that they will be automagically generated leveraging Python 3.6 typing to handle argument generation in the future.

```py
class MyModel(Base):
   # ... fields

   def create(self, foo: String, bar:Optional[String] = None):
      # ...

# would imply

class CreateMyModel(graphene.Mutation):
   class Arguments:
      foo = graphene.String(required=True)
      bar = graphene.String()

   myModel = # node for my model

   def mutate(self, info, **kwargs):
      return CreateMyModel(myModel=MyModel.create(**kwargs))
```

Issue #209, opened 2019-04-25
Discussion: General Mechanism to Selectively Override Default Behavior


jnak:
### Overview

In order to convert SQLAlchemy object models (SAM) into a full-blown GraphQL schema, SQLAlchemyObjectType (SOAT) does a lot of automatic mapping and conversions:
- it automatically names Graphene fields after their corresponding SAM column names   
- it automatically converts SAM column types to common Graphene types 
- it automatically converts SAM relationships to Relay-style paginated fields
- it automatically uses SAM descriptions to populate Graphene field's descriptions 
- it automatically converts SQLAlchemy enums to Graphene enums (see https://github.com/graphql-python/graphene-sqlalchemy/issues/208) 

While it is convenient that SOAT does all this out-of-the-box, developers should ultimately be in control of the Graphene types generated. Currently we can override SOAT default's behavior in 2 ways.

**1. Use `Meta`** 

The `Meta` of SOAT allows the overriding of a few select automatic behaviors. For example, we can restrict what columns are exposed via the `only_fields` and `exclude_fields` parameters. In this example, `Meta` works fine because those parameters essentially act as toggles that completely disable or enable SOAM behavior for each field.

Similarly, we can currently override SOAT default connection field factory via `connection_field_factory`. But this is not flexible enough because we may want to have SOAT use different factories for different fields. For example, it may make sense to have only one paginated field be sortable (hence use `SQLAlchemyConnectionField`) and all the other fields be regular Relay-style collections (hence use `UnsortedSQLAlchemyConnectionField`).

One simple workaround would be to have a have a `connection_field_factories` parameter that takes a dictionary of field names to factories:
```
class MyType(SQLAlchemyObjectType):
    class Meta:
         model = MyModel
         default_connection_field_factory = MyConnectionField
         connection_field_factories = {
             'my_connection_field': SortedConnectionField,
         }
```

While that works fine, it leads to the creation of many parameters in the `Meta` class (eg https://github.com/graphql-python/graphene-sqlalchemy/pull/178 and https://github.com/graphql-python/graphene-sqlalchemy/issues/208#issuecomment-485819664) and we may end up with a `Meta` class that becomes hard to maintain.

My main concern though with this pattern is that scatters the logic for a field in separate `Meta` parameters. For example, we would need to do something like to rename a field, specify a connection factory and deprecate it:
```
class MyType(SQLAlchemyObjectType):
    class Meta:
         model = MyModel
         connection_field_factories = {
[...]

Issue #208, opened 2019-04-18
Discussion: Auto-creation of Graphene Enums


Cito:
In PR #98 I made an attempt to improve the auto generation of Graphene Enums, but it turned out that this feature still needs some discussion. Let's clarify here what we really want to achieve first.

In the following:
* g-enum = Graphene Enum types
* sa-enum = SQLAlchemy Enum type
* py-enum = Python Enum type
* sql-enum = Database enum data type

The sql-enums are supported as actual data types by some databases, otherwise implemented using name constraints. Note that both g-enums and sa-enums can be based either on py-enums or on simple lists of values. 

For the examples below, I use the following imports:

```
from enum import Enum as PyEnum
import sqlalchemy as sa
from sqlalchemy.ext.declarative import declarative_base

Base = sa.ext.declarative.declarative_base()
```

When g-enums are created from sa-enums by `graphene_sqlalchemy`, they must be given a GraphQL type name. We must decide how these name are generated.

### 1) sa-enums based on py-enums

Example 1:

```
class PetKind(PyEnum):
    cat = 1
    dog = 2

class Pet(Base):
    __tablename__ = "pets"
    name = sa.Column(sa.String(30), primary_key=True)
    kind = sa.Column(sa.Enum(PetKind))
```

I think it's clear that in this case, the g-enum should be derived from `PetKind` and thus get the same name `PetKind`. Since GraphQL and Python have the same conventions for type/class names, this case is all good and fine.

### 1a) sa-enums based on py-enums, with SQL name

Example 1a:

```
class Pet(Base):
    __tablename__ = "pets"
    name = sa.Column(sa.String(30), primary_key=True)
    kind = sa.Column(sa.Enum(PetKind, name='kind_of_pet'))
```

The sa-enums can be declared with a name argument which is used for name constraints or as sql-enum type name on the database. In this case, I think the name of the py-enum should take precedence for us; the generated g-enum should still be derived from `PetKind`, and thus have the name `PetKind`, not `KindOfPet`.

### 2) sa-enums based on values, with SQL name

Example 2:

```
class Pet(Base):
    __tablename__ = "pets"
    name = sa.Column(sa.String(30), primary_key=True)
    kind = sa.Column(sa.Enum('cat', 'dog', name='kind_of_pet'))
```

Note: The name here is passed as argument to Enum, not Column. I.e. the sql-enum name is `kind_of_pet`, while the column name is `kind`.  This case is interesting. 

[...]

Issue #29, opened 2016-12-28
Auto generated mutations given model


aminghadersohi:
Is there anything on the roadmap to create a class similar to SQLAlchemyObjectType but for mutations that would take a model in its Input and automatically create a graphene.Mutation subclass that has all the fields in it? The reason I ask is that I am thinking about brewing my own and if its not on the roadmap I'd be happy to contribute unless there is a good reason why this is bad. I think I can create a SQLAlchemyMutationMeta and SQLAlchemyMutation class to take the model from the Input class, fetch the fields in it and pass appropriate attrs to MutationMeta.

marcovc:
Do you have any code to share for doing that? I just realized that graphene-sqlalchemy is completely useless to me if I have to duplicate my model definitions anyway (because of the mutations)... 
Thanks

aminghadersohi:
The SQLAlchemy Model/Store:

```
Base = sqlalchemy.ext.declarative.declarative_base()

class ClientRoleStore(Base):
    __tablename__ = "clientrole"
    uuid = Column(CHAR(36), primary_key=True, nullable=False, default=lambda: str(uuid1()))
    created_at = Column('created_at', DateTime, nullable=False,
                        server_default=func.now())
    updated_at = Column('updated_at', DateTime, index=True, nullable=False,
                        server_onupdate=func.now(),
                        server_default=func.now())
    deleted_at = Column('deleted_at', DateTime, nullable=True)
    name = Column(String(150), unique=True, nullable=False)
    description = Column(String(500))
    enabled = Column(Boolean, nullable=False, default=False)
```

The GraphQL object type: 

```
class ClientRole(SQLAlchemyObjectType):
    class Meta:
        model = ClientRoleStore
        interfaces = (relay.Node,)
```

This is the main goods. Extend MutationMeta:
```
# just a util function
def camel_to_snake(s):
    s = re.sub('(.)([A-Z][a-z]+)', r'\1_\2', s)
    s = re.sub('(.)([0-9]+)', r'\1_\2', s)
    return re.sub('([a-z0-9])([A-Z])', r'\1_\2', s).lower()

class SQLAlchemyMutationMeta(MutationMeta):
    @staticmethod
    def __new__(cls, name, bases, attrs):
        # Also ensure initialization is only performed for subclasses of
        # SQLAlchemyMutation
        if not is_base_type(bases, SQLAlchemyMutationMeta):
            return type.__new__(cls, name, bases, attrs)

[...]

PR #213, opened 2019-05-14
Adding SQLAlchemyInputObjectType

Rafik-Belkadi:


jnak:
Hello @Rafik-Belkadi, thanks for sending a PR.  I agree with you that there is potential for `graphene-sqlalchemy` to help with mutations. That said, it's not entirely clear how you intend to use your change since you haven't added any tests.

Since it's a whole new functionality, I suggest that before adding tests you start a separate issue / conversation that outlines what your goals are, the different strategies to go about this and your proposed solution (see https://github.com/graphql-python/graphene-sqlalchemy/issues/212#issuecomment-490950853). You can see examples of recent successful design conversations [here](https://github.com/graphql-python/graphene-sqlalchemy/issues/208) and [here](https://github.com/graphql-python/graphene-sqlalchemy/issues/209). These conversations are great way to gather early feedback and reach consensus before you start get into the details of the implementation. Obviously, if the problem you are solving is straightforward, the conversation does not need to be as long as these two examples.

Here are some of the considerations that I think we should discuss as a group:
- Why not use use a the already defined SQLAlchemyObjectType types in the mutation?
```py
class ShipMutation(relay.ClientIDMutation):
    class Input:
        ship = graphene.Field(Ship)
    ship = graphene.Field(Ship)
```
- Should the input include connection relationships to other types (as it is in the current version of this PR)? If not, how do want to handle those?
- Does it make sense to include hybrid properties, composite columns and [column properties](https://docs.sqlalchemy.org/en/13/orm/mapping_columns.html#sqlalchemy.orm.column_property)?

Cheers
J

sreerajkksd:
There are some discussions about the mutation support in https://github.com/graphql-python/graphene-sqlalchemy/issues/29

sreerajkksd:
@jnak that might not work.

```
class ShipMutation(relay.ClientIDMutation):
    class Input:
        ship = graphene.Field(Ship)
    ship = graphene.Field(Ship)
```

`ShipMutationInput.ship` field should be of `Input` type.  `Ship` in your example is a subclass of `SQLAlchemyObjectType`.

Or, Am I missing anything ?

himat:
I'm getting this error while trying to use this code
`TypeError: convert_sqlalchemy_column() missing 1 required positional argument: 'resolver'`

baderj:
```
class ShipMutation(graphene.Mutation):
    class Input:
        ship = graphene.Field(Ship)
    ship = graphene.Field(Ship)
```

This fails with ``ValueError: Expected input to be Argument, but received Field``. Currently the only way I found is use the ``graphene.InputObjectType`` and specify each column again from the sqlalchemy model. 

Is there a reason, why ``SQLAlchemyInputObjectType`` was never merged? Having a nice

```
class Ship(SQLAlchememyObjectType):
    class Meta:
        model = tables.Ship
```

followed by the verbose

```
class ShipInput(InputObjectType):
    name = graphene.String()
    registration = graphene.String()
    height = graphene.Number()
    width = graphene.Number()
    ... 50 other columns copied from orm

```
Is not only very annoying to type out, it also look very inconsistent. I don't really need the ``SQLAlchemyInputObjectType`` if there were a way to use an ``SQLAlchemyObjectType`` as ``Arguments`` or ``Input`` to the ``graphene.Mutation``
 

```

seanxlliu:
@jnak  Thanks for making this great library! Sorry for bringing back this old question again. I have seen many ppl request this feature
#212 #216  #285 
Now I have encoutered the exactly same issue in my project. We have a huge mount of data models need mutations though graphql APIs. I have read your comment above, however, the types cannot be reused as inputs. [here](url) is the philosophy of GraphQL. Long in short:

> The GraphQL Object type (ObjectTypeDefinition)... is inappropriate for re‐use [as an input], because Object types can contain fields that define arguments or contain references to interfaces and unions, neither of which is appropriate for use as an input argument. For this reason, input objects have a separate type in the system.

So, what is the next move for the improvement so that we can keep working on?

seanxlliu:
More from [GraphQL spec](https://spec.graphql.org/June2018/#sec-Input-and-Output-Types):

> IsInputType(type)
If type is a List type or Non‐Null type:
  Let unwrappedType be the unwrapped type of type.
  Return IsInputType(unwrappedType)
If type is a Scalar, Enum, or Input Object type:
  Return true
Return false

IsOutputType(type)
If type is a List type or Non‐Null type:
  Let unwrappedType be the unwrapped type of type.
  Return IsOutputType(unwrappedType)
If type is a Scalar, Object, Interface, Union, or Enum type:
  Return true
Return false

PR #214, opened 2019-05-14
Implement Mechanism to Selectively Override Automatic Field Creations

jnak:
This implements the overriding mechanism discussed in https://github.com/graphql-python/graphene-sqlalchemy/issues/209.

Changes include:
- Add `ORMField` class
- The main overridable parameters are: `type`, `description`, `deprecation_reason` and `required`
- We can name fields differently using `prop_name`. This was preferred over `name` to avoid confusion / collision with `graphene.Field` parameters.
- Add tests for all types of SQLAlchemy properties: columns, relationships, column properties, hybrid properties and and composite properties. 
- Cleanups and re-organize some tests. 

TODOs that I'm planning to address as soon as this PR is merged: 
- Add a way to override relationship types. 
- Add a way to override composite fields default parameters. 
- Add example and documentation


graphene_sqlalchemy/fields.py
```diff
@@ -97,22 +97,22 @@ def __init__(self, type, *args, **kwargs):
...
 
 
-def createConnectionField(_type):
+def createConnectionField(_type, **field_kwargs):
     log.warning(
         'createConnectionField is deprecated and will be removed in the next '
         'major version. Use SQLAlchemyObjectType.Meta.connection_field_factory instead.'
     )
-    return __connectionFactory(_type)
+    return __connectionFactory(_type, **field_kwargs)
```
jnak: FYI There is a chance that people will have to update their registered factory if they want to start using `ORMField`. But this a very likely because `__connectionFactory` is usually set to a field class.

graphene_sqlalchemy/tests/test_schema.py
```diff
@@ -1,50 +0,0 @@
-from py.test import raises
```
jnak: FYI I moved those tests to test_types.py because it was not clear where to add new tests.

graphene_sqlalchemy/types.py
```diff
@@ -21,70 +24,91 @@
...
+        common_kwargs = {kwarg: value for kwarg, value in common_kwargs.items() if value is not None}
+        self.kwargs = field_kwargs
+        self.kwargs.update(common_kwargs)
+
+
 def construct_fields(
     obj_type, model, registry, only_fields, exclude_fields, connection_field_factory
 ):
     inspected_model = sqlalchemyinspect(model)
+    all_model_props = OrderedDict(
```
wyattanderson: This seems inefficient and potentially inflexible. Could we use [Mapper.attrs](https://docs.sqlalchemy.org/en/13/orm/mapping_api.html#sqlalchemy.orm.mapper.Mapper.attrs) or [Mapper.all_orm_descriptors](https://docs.sqlalchemy.org/en/13/orm/mapping_api.html#sqlalchemy.orm.mapper.Mapper.all_orm_descriptors) instead of combining all of these, which I think are subsets of the aforementioned?

graphene_sqlalchemy/types.py
```diff
@@ -21,70 +24,91 @@
...
+        inspected_model.column_attrs.items() +
+        inspected_model.composites.items() +
+        [(name, item) for name, item in inspected_model.all_orm_descriptors.items()
+            if isinstance(item, hybrid_property)] +
+        inspected_model.relationships.items()
+    )
+
+    auto_orm_field_names = []
+    for prop_name, prop in all_model_props.items():
```
wyattanderson: `iteritems`

graphene_sqlalchemy/types.py
```diff
@@ -21,70 +24,91 @@
...
+        if (only_fields and prop_name not in only_fields) or (prop_name in exclude_fields):
+            continue
+        auto_orm_field_names.append(prop_name)
+
+    # TODO Get ORMField fields defined on parent classes
```
wyattanderson: For the `TODO`, maybe something like this would be helpful:

```
itertools.chain.from_iterable(base.__dict__ for base in obj_type.__mro__)
```

graphene_sqlalchemy/types.py
```diff
@@ -21,70 +24,91 @@
...
+    custom_orm_fields_items = []
+    for attname, value in list(obj_type.__dict__.items()):
```
wyattanderson: `.items()` is already a list; the `list()` is unnecessary. I'd consider using `iteritems()`, though. You could also build this up as a comprehension:

```
custom_orm_fields_items = [
    attname, value
    for attname, value in obj_type.__dict__.iteritems()
    if isinstance(value, ORMField)
]
```

graphene_sqlalchemy/types.py
```diff
@@ -21,70 +24,91 @@
 from .utils import get_query, is_mapped_class, is_mapped_instance
 
 
+class ORMField(OrderedType):
+    def __init__(
+        self,
+        type=None,
+        prop_name=None,
```
wyattanderson: The underlying name for this on the SQLAlchemy `ColumnProperty` is `key`; I think it might be nice to avoid coming up with new names where possible and calling this `key` instead of `prop_name`. What do you think?

graphene_sqlalchemy/types.py
```diff
@@ -21,70 +24,91 @@
...
+           'description': description,
+           'deprecation_reason': deprecation_reason,
+           'required': required,
+        }
+        common_kwargs = {kwarg: value for kwarg, value in common_kwargs.items() if value is not None}
+        self.kwargs = field_kwargs
+        self.kwargs.update(common_kwargs)
+
+
 def construct_fields(
```
wyattanderson: I think this could benefit from—at the least—a docstring describing the overall flow and the intent. Even something that outlines the logic:

- gather all relevant attributes from the model
- gather all custom ORM fields defined on the type
- merge in overrides and build up all the fields

This logic is pretty dense and I think docstrings and inline comments would be beneficial to future readers.

graphene_sqlalchemy/types.py
wyattanderson: How about `property` or `attribute`? Marshmallow uses `attribute` for the same purpose https://marshmallow.readthedocs.io/en/3.0/api_reference.html#module-marshmallow.fields

graphene_sqlalchemy/types.py
wyattanderson: Fair enough.

graphene_sqlalchemy/types.py
wyattanderson: How so? Works for me locally.

graphene_sqlalchemy/types.py
jnak: They seem both fine to me but I think prefer `property` because it seems a bit more consistent with `sqlalchemy`. Also I don't think we should base our judgement too much on what  `marshmallow` does since it is not directly related to `graphene` or `sqlalchemy`.

The more I think about it the more I like the idea of prefixing the param name with `orm_` because it is consistent with how `ORMField` relates to `graphen.Field`. It also completely avoids any risk of collision with `graphene` parameters down the road. 

How does `orm_property` sound to you? If it's too long we can shorten it to `orm_prop`.

wyattanderson: I think `attribute` is a pretty well-known name both in Python itself (`__getattr__`, `getattr`), the library ecosystem (Marshmallow, [attrs](https://www.attrs.org/en/stable/)) or in SQLAlchemy itself (`InstrumentedAttribute`, `Mapper.column_attrs`, it's all over [the class mapping API](https://docs.sqlalchemy.org/en/13/orm/mapping_api.html), in which case `attribute` might be more consistent with SQLAlchemy than `property`). We even use `attribute` in the docstring you wrote here.

Another option could be `field`, since we already have `exclude_fields` and `only_fields`.

Or `name`, like how SQLAlchemy `Column()` takes an explicit `name` of the column you're creating a mapping for, letting you name the field on your model differently than the underlying column in the DB.

I think the `orm_` prefix is unnecessarily redundant and unlikely to collide with parameters down the line.

jnak: As discussed offline, I renamed to `model_attr`.

graphene_sqlalchemy/tests/test_query.py
```diff
@@ -101,29 +112,23 @@ class Meta:
...
 
         def resolve_reporter(self, _info):
             return session.query(Reporter).first()
 
-        def resolve_article(self, _info):
-            return session.query(Article).first()
-
     query = """
-        query ReporterQuery {
+        query {
```
Nabellaleen: Is there a reason to "de-name" the query here ?

graphene_sqlalchemy/tests/test_types.py
```diff
@@ -200,42 +330,34 @@ class _TestSQLAlchemyConnectionField(SQLAlchemyConnectionField):
 
 
 def test_default_connection_field_factory():
-    _registry = Registry()
```
Nabellaleen: Why ? ;)

graphene_sqlalchemy/tests/test_query.py
jnak: There was no reason to name the query so I removed to make tests more consistent and simple.

graphene_sqlalchemy/tests/test_types.py
jnak: I realized that was not needed because the global registry is reset between tests (see [`confest.py`](https://github.com/graphql-python/graphene-sqlalchemy/blob/8ad1f75639ced6b00688b63c1a7b4a6a1acdf5af/graphene_sqlalchemy/tests/conftest.py#L13)).

Issue #216, opened 2019-05-22
Bug "type object 'Employee' has no attribute 'sort_argument'"


royhananwar:
I just clone this repo and try flask example
but when I try to run it without have any changing code
I got an error like this

`Traceback (most recent call last):
    File "./app.py", line 5, in <module>
    from schema import schema
  File "/home/examples/flask_sqlalchemy/schema.py", line 28, in <module>
    class Query(graphene.ObjectType):
  File "/home/examples/flask_sqlalchemy/schema.py", line 32, in Query
    Employee, sort=Employee.sort_argument())
AttributeError: type object 'Employee' has no attribute 'sort_argument'`

What should I do?

Thank you!

Cito:
Are you sure you have the version from the cloned repo installed when running the app, not an older version?

fatore:
I got the same here while following the steps from https://github.com/graphql-python/graphene-sqlalchemy/tree/master/examples/flask_sqlalchemy. 

I assumed the proper versions were defined in the [`requirements.txt`](https://github.com/graphql-python/graphene-sqlalchemy/blob/master/examples/flask_sqlalchemy/requirements.txt) file.

Cito:
That `requirements.txt` file installs graphene-sqlalchemy from PyPI which is not compatible with the current development version. I think the `requirements.txt` file should be replaced with this:

```
-e ../../
Flask-GraphQL
```

That will install the currently checked out version. Graphene and SQLAlchemy will be automatically installed as dependency in a matching version. Only Flask-GraphQL needs to be installed explicitly.

Can you remove the virtualenv directory `env` and try again with that `requirements.txt` file?

jnak:
Good point. There you go https://github.com/graphql-python/graphene-sqlalchemy/pull/218

royhananwar:
> That `requirements.txt` file installs graphene-sqlalchemy from PyPI which is not compatible with the current development version. I think the `requirements.txt` file should be replaced with this:
> 
> ```
> -e ../../
> Flask-GraphQL
> ```
> 
> That will install the currently checked out version. Graphene and SQLAlchemy will be automatically installed as dependencies with matching versions. Only Flask-GraphQL needs to be installed explicitly and needs to be added in `requirements.txt`.
> 
> Can you remove the virtualenv directory `env` and try again with that `requirements.txt` file?

Yes, you are right. I need to use the old `requirements.txt`
Thank you!
it worked like charms!

PR #217, opened 2019-05-23
Release version 2.2.0

jnak:
# Changelog

- Improve creation of column and sort enums (https://github.com/graphql-python/graphene-sqlalchemy/commit/db3e9f4c3baad3e62c113d4a9ddd2e3983d324f2)

### DIFF
--- a/graphene_sqlalchemy/__init__.py
+++ b/graphene_sqlalchemy/__init__.py
@@ -2,7 +2,7 @@ from .types import SQLAlchemyObjectType
 from .fields import SQLAlchemyConnectionField
 from .utils import get_query, get_session
 
-__version__ = "2.1.2"
+__version__ = "2.2.0"
 
 __all__ = [
     "__version__",


jnak:
@Cito This is releasing your enum PR. This will make it easier for people [to follow the enum example](https://github.com/graphql-python/graphene-sqlalchemy/issues/216)

Cito:
@jnak - I think we should also change the `requirements.txt` file like I sugested in #216. Since the ReadMe in the example directory recommends cloning from GitHub, we should install the currently cloned version instead of the version released on PyPI. Also the current `requirements.txt` file has a version restriction for SQLAlchemy that is incompatible with the current restriction in our setup.py, so it needs to be fixed anyway.

jnak:
@Cito Now that https://github.com/graphql-python/graphene-sqlalchemy/pull/218 is merged, can you also accept this one so people can start using your change? Some issues were opened that might be not be happening with your changes (ex https://github.com/graphql-python/graphene-sqlalchemy/issues/219).

PR #218, opened 2019-05-23
Fix Example Apps

jnak:
- Update requirements.txt so they point at the local version of `graphene-sqlalchemy`
- Use absolute imports in `nameko_sqlalchemy` because `./run.sh` does not support absolute imports


jnak:
@Cito Sure. You have to re-approve the PR

Issue #228, opened 2019-06-13
Documentation on connection_field_factory


rdemetrescu:
I'm using graphene-sqlalchemy version 2.2.0 and I'm getting lots of `createConnectionField is deprecated and will be removed in the next major version. Use SQLAlchemyObjectType.Meta.connection_field_factory instead.` warnings.

Is there any example on how to configure this connection_field_factory? Are there any default factories?  So far I couldn't find any example using this new feature on the project site.

Thank you!

zmwangx:
The problem here is that `createConnectionField` is used in `default_connection_field_factory`, so the library itself is relying on its supposedly deprecated API, and users get a WARNING through no fault of their own.

vaskokj:
Same exact issue. 

Whats odd to me is that it happens when you add more than one class that uses SQLAlchemyObjectType.

With only one class subclassing the SQLAlchemyObjectType you don't get the error. If you uncomment the commented out section you will get the warning. 

```
import graphene
from graphene import relay
from graphene_sqlalchemy import SQLAlchemyObjectType, SQLAlchemyConnectionField
from myapp.models import Channel as ChannelModel, Model as ModelModel

# class Model(SQLAlchemyObjectType):
#     class Meta:
#         model = ModelModel
#         interfaces = (relay.Node, )

class Channel(SQLAlchemyObjectType):
    class Meta:
        model = ChannelModel
        interfaces = (relay.Node, )

class ChannelConnection(relay.Connection):
    class Meta:
        node = Channel

class Query(graphene.ObjectType):
    node = relay.Node.Field()
    all_channels = SQLAlchemyConnectionField(ChannelConnection)

schema = graphene.Schema(query=Query)
```

zmwangx:
> With only one class subclassing the SQLAlchemyObjectType you don't get the warning.

Nah, that’s probably because your `Channel` does not auto-generate a connection field, while your `Model` does. Check your schema, including nested levels. There’s nothing magical about the warning: it’s printed every time a connection field is created with the default factory.

vaskokj:
@zmwangx I added the ConnectionField to the Channel and it still does it. Uncomment the comments and it will throw the error.

```
import graphene
from graphene import relay
from graphene_sqlalchemy import SQLAlchemyObjectType, SQLAlchemyConnectionField
from myapp.models import Channel as ChannelModel, Model as ModelModel


#class Model(SQLAlchemyObjectType):
#   class Meta:
[...]

PR #229, opened 2019-06-15
Fix `createConnectionField` deprecation warnings

jnak:
- Fixes https://github.com/graphql-python/graphene-sqlalchemy/issues/228. These deprecation warnings were incorrectly raised by `default_connection_field_factory`.

- Releases fix as `2.2.1`

### DIFF
--- a/graphene_sqlalchemy/tests/test_types.py
+++ b/graphene_sqlalchemy/tests/test_types.py
@@ -364,33 +364,35 @@ def test_custom_connection_field_factory():
 
 
 def test_deprecated_registerConnectionFieldFactory():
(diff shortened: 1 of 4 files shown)


graphene_sqlalchemy/tests/test_types.py
```diff
@@ -224,6 +224,19 @@ class Meta:
     assert pets_field.type().description == 'Overridden'
 
 
+def test_invalid_model_attr():
```
jnak: FYI I had to add an unrelated test because `coveralls` thought incorrectly that the overall coverage went down without this.

Issue #122, opened 2018-03-27
Invalid SQLAlchemy Model


rvandyk:
Hey guys I hope someone can help me on this, i'm really stuck....

database automaping:
```
engine = create_engine(" blablabla", convert_unicode=True)

db_session = scoped_session(sessionmaker(autocommit=False,
                                         autoflush=False,
                                         bind=engine))

Base = automap_base()
Base.prepare(engine, reflect=True, generate_relationship=_gen_relationship, name_for_scalar_relationship = name_for_scalar_relationship, classname_for_table=camelize_classname, name_for_collection_relationship=pluralize_collection)

Buyer = Base.classes.Buyer

Base.query = db_session.query_property()
```

schema : 
```
import graphene
import json
from graphene import relay
from graphene_sqlalchemy import SQLAlchemyConnectionField, SQLAlchemyObjectType

from database import Buyer as BuyerModel

class Buyer(SQLAlchemyObjectType):

    class Meta:
        model = BuyerModel
        interfaces = (relay.Node, )

class Query(graphene.ObjectType):
    node = relay.Node.Field()
    all_buyers = SQLAlchemyConnectionField(Buyer)

schema = graphene.Schema(query=Query)
```

Produces the error : 
`AssertionError: You need to pass a valid SQLAlchemy Model in Buyer.Meta, received "<class 'sqlalchemy.ext.automap.Buyer'>".`

Fedalto:
There's probably something wrong with your SQLAlchemy models, and nothing to do with the graphene schema.
I recently opened #121 explaining a bug that graphene-sqlalchemy does not show you the actual error. Your can try to put a breakpoint in `is_mapped_class` to see the exception SQLAlchemy is throwing.

AndreCAndersen:
Thanks @Fedalto, I tried to dive in. Ultimately all `is_mapped_class(model)` does is call [`def manager_of_class(cls)`](https://github.com/zzzeek/sqlalchemy/blob/8f7766cc61479f3c9220c640230eeecd3d49ccc8/lib/sqlalchemy/orm/base.py#L208), in the `orm/base.py` module, using your defined model. It's a tiny function with does the following:

    def manager_of_class(cls):
        return cls.__dict__.get(DEFAULT_MANAGER_ATTR, None)

... where `DEFAULT_MANAGER_ATTR = '_sa_class_manager'`.

In essence it requires that you're model has a field '_sa_class_manager' if not it returns `None` then propagates the error up.

[...]

PR #238, opened 2019-07-22
Removed assertions that were hiding exceptions

zefciu:
The two assertions that were testing the sqlalchemy classes and instances
worked by catching and silencing exceptions. This could cause debugging
problems like:

https://github.com/graphql-python/graphene-sqlalchemy/issues/121
https://github.com/graphql-python/graphene-sqlalchemy/issues/122

on the other hand without them, if someone uses unmapped class/model it
would be trivial to diagnose the problem. So I believe we would be
better off without them.

Fixes: #121

### DIFF
--- a/graphene_sqlalchemy/types.py
+++ b/graphene_sqlalchemy/types.py
@@ -21,7 +21,7 @@ from .enums import (enum_for_field, sort_argument_for_object_type,
                     sort_enum_for_object_type)
 from .fields import default_connection_field_factory
 from .registry import Registry, get_global_registry
-from .utils import get_query, is_mapped_class, is_mapped_instance
+from .utils import get_query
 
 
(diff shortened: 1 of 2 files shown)


jnak:
Thanks for sending a PR. I agree it's frustrating that SQLAlchemy issues are currently swallowed. But I also  think there is value in providing helpful error messages such as the ones from the failing tests. Would you updating this PR so it does both and both cases are tested? 

Thanks!

ryarasi:
Why is this an open issue even after 2 yrs??! 

So many of us are wasting countless hours on troubleshooting with invisible debug data :'(

rimu:
This project is pretty much maintenance-only. Better not to use it.

erikwrede:
Closing this in favor of #169

PR #240, opened 2019-08-05
add support for enums in sqlalchemy ChoiceType

metheoryt:
Hi! I've found that I cannot use python enums with sqlalchemy_utils's ChoiceType, so here's my solution for it

### DIFF
--- a/graphene_sqlalchemy/converter.py
+++ b/graphene_sqlalchemy/converter.py
@@ -7,6 +7,7 @@ from graphene import (ID, Boolean, Dynamic, Enum, Field, Float, Int, List,
                       String)
 from graphene.types.json import JSONString
 
+from enum import EnumMeta
 from .enums import enum_for_sa_enum
 from .registry import get_global_registry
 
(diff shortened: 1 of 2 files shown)


jnak:
Thanks for sending a PR. Would you mind updating the tests in `test_converter.py` so both cases are tested? 

You can see how to run the tests locally here: https://github.com/graphql-python/graphene-sqlalchemy/blob/master/CONTRIBUTING.md

graphene_sqlalchemy/converter.py
```diff
@@ -159,11 +161,15 @@ def convert_enum_to_enum(type, column, registry=None):
     return lambda: enum_for_sa_enum(type, registry or get_global_registry())
 
 
-# TODO Make ChoiceType conversion consistent with other enums
```
jnak: I think we should still plan to convert ChoiceType enums like other we do with other SQLAlchemy enums column types. That is:
- we should follow the same [consistent naming scheme](https://github.com/graphql-python/graphene-sqlalchemy/blob/9a0f7400957c5b5f5f8e952db2502ef1f79fac1f/graphene_sqlalchemy/enums.py#L9-L42)
- we should [register it in the registry](https://github.com/graphql-python/graphene-sqlalchemy/blob/9a0f7400957c5b5f5f8e952db2502ef1f79fac1f/graphene_sqlalchemy/enums.py#L45-L55) so we can later access it (see [example](https://github.com/graphql-python/graphene-sqlalchemy/blob/db3e9f4c3baad3e62c113d4a9ddd2e3983d324f2/graphene_sqlalchemy/tests/test_query_enums.py#L23-L24))

Would you mind leaving the TODO?

jnak:
@metheoryt Btw in case you're interested, here is more context on the TODO. The enums auto-generated by ChoiceLabel are not as clean and robust as other enum columns:
- the name of the enum include the the table and the column name
- the name of the enum cannot be customized
- a duplicated enum is created (with a different name) if the ChoiceType is shared across multiple columns

This [PR](https://github.com/graphql-python/graphene-sqlalchemy/pull/210) added proper support for enums but did not include ChoiceType. I think it would be quite straightforward to have ChoiceType leverage the same infrastructure. 

Cheers

PR #244, opened 2019-08-21
Check against six.string_types instead of str for python 2 compatibility

aubustou:
If you named your columns with Unicode in Python 2, it obviously won't work...
In order to keep Python 2 and 3 compatibilities, better to use six.string_types.

### DIFF
--- a/graphene_sqlalchemy/enums.py
+++ b/graphene_sqlalchemy/enums.py
@@ -1,3 +1,4 @@
+import six
 from sqlalchemy.orm import ColumnProperty
 from sqlalchemy.types import Enum as SQLAlchemyEnumType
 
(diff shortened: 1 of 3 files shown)


jnak:
Makes sense.

It looks like you have a minor [lint issue](https://travis-ci.org/graphql-python/graphene-sqlalchemy/jobs/574899453#L281-L288). You should be able to fix it automatically by running `tox -e pre-commit` (see [how to set up your local dev env](https://github.com/graphql-python/graphene-sqlalchemy/blob/master/CONTRIBUTING.md#local-development))

Also would you mind adding or updating a test case to prevent regressions in the future?

Let me know if you run into any issue.
Thanks

aubustou:
Once i manage to install that damned Python 3.7 on my install, I'll fix it.

PR #249, opened 2019-10-17
Add database session to the example

clemens-tolboom:
Coming from https://docs.graphene-python.org/projects/sqlalchemy/en/latest/tutorial/ as a python noob I failed to run their example but could fix this example by adding the database session.


README.md
```diff
@@ -58,6 +58,18 @@ class Query(graphene.ObjectType):
 schema = graphene.Schema(query=Query)
 ```
 
+We need a database session first:
```
erikwrede, line 61: I would move this up to the "Here is a simple SQLAlchemy model:" part and add 
```python
# We will need this for querying
Base.query = db_session.query_property()
```
from the docs, since that is essential to using graphene-sqa.

erikwrede (changes requested):

erikwrede (approved):

erikwrede:
Better late than never - thanks for your contribution 😊

PR #250, opened 2019-10-17
ValueError: The options 'only_fields' and 'exclude_fields' cannot be …

clemens-tolboom:
…both set on the same type.

Not sure what to choose I removed both as these do not help for the example imho.

See https://github.com/graphql-python/graphene-sqlalchemy/blob/0544f812b37f8e6b49ed3363e7010c08a600be1d/graphene_sqlalchemy/types.py#L237

### DIFF
--- a/README.md
+++ b/README.md
@@ -43,6 +43,10 @@ from graphene_sqlalchemy import SQLAlchemyObjectType
 class User(SQLAlchemyObjectType):
     class Meta:
         model = UserModel
+        # use `only_fields` to only expose specific fields ie "name"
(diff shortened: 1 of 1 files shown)


jnak:
Hello, thanks for catching this and sending a PR. I feel it's important that `only_fields` and `exclude_fields` are documented in the `README`. How about we keep the `exclude_fields` declaration since it tends to be the most common and adding a comment saying that 'only_fields` can be used to whitelist fields?

clemens-tolboom:
I've rephrased the PR. Hope that's better :)

PR #253, opened 2019-10-29
Fix N+1 problem for one-to-one and many-to-one relationships

jnak:
Hi everyone,

This PR fixes the N+1 problem for `one-to-one` and `many-to-one` relationships. If we think this approach makes sense, I'll send a separate PR to address `one-to-many` and `many-to-many` relationships.

**Design goals**
- Be completely seamless to end-users. You should not need to modify your `SQLAlchemy` models or your `SQLAlchemyObjectType` models to benefit from this.

- Work with any `SQLAlchemy` relationships, including ones that have complex `primaryjoin` and `secondaryjoin` conditions.

- Avoid generating one large SQL query using nested `JOIN` in the resolver of the root field of the query. For non-trivial queries, the joined SQL statement is complex with hard-to-predict performances. See SQLAlchemy's [comparison of `selectin` with `joined` loading](https://docs.sqlalchemy.org/en/13/orm/loading_relationships.html#select-in-loading) for more context. 

- Fully work in schemas that include non-`SQLAlchemyObjectType` Graphene types. This batching optimization should compose well with other types, including those that use `Dataloader` under the hood.

- Avoid building SQL queries manually. It's non-trivial to generate correct SQL statements that support complex joins and composite primary keys across a variety of DBs. We should lean as much as possible on SQLAlchemy to do the heavy lifting.

**Non-goals**
- Only select the columns / fields that are being queried. This will be addressed in a separate PR.

- Caching records so we don't fetch the same record twice from the DB. This is tricky to get right. We can revisit this later if there is a need.  

**Solution**

The idea here is to have `SQLAlchemyObjectType` automatically generate batch resolvers for relationships. The resolvers use [`Dataloader`](https://docs.graphene-python.org/en/latest/execution/dataloader/) to collect all the similar `relationship` property accesses. From there, we ask `SQLAlchemy` to load all the relationships of those parents as one SQL query.

`SQLAlchemy` does not have a public API to do this but we can piggyback on some internal APIs of the [`selectin`](https://docs.sqlalchemy.org/en/13/orm/loading_relationships.html#select-in-loading) eager loading strategy. It's a bit hacky but I think it's better than re-implementing and maintaining a big chunk of the [`selectin` loader logic](https://github.com/sqlalchemy/sqlalchemy/blob/master/lib/sqlalchemy/orm/strategies.py#L2294-L2481) ourselves. You can find more details about this approach in the docstring of the batch loader. Hopefully we can get some feedback from the `SQLAlchemy` maintainers (cc @zzzeek). 

If we decide to move forward with this PR, I'll make sure to add more test cases so we can catch regressions in the event the internal `SQLALchemy` APIs we rely on change.   

Cheers,
J

### DIFF
--- a/graphene_sqlalchemy/types.py
+++ b/graphene_sqlalchemy/types.py
@@ -8,7 +8,6 @@ from sqlalchemy.orm import (ColumnProperty, CompositeProperty,
 from sqlalchemy.orm.exc import NoResultFound
 from sqlalchemy.orm.query import QueryContext
 from sqlalchemy.orm.strategies import SelectInLoader
-from sqlalchemy.orm.util import PathRegistry
 
 from graphene import Field
 from graphene.relay import Connection, Node
@@ -265,9 +264,6 @@ def _get_relationship_resolver(obj_type, relationship_prop, model_attr):
 
             loader = SelectInLoader(relationship_prop, (('lazy', 'selectin'),))
 
-            # The path is a fixed single token in this case
-            path = PathRegistry.root + parent_mapper._path_registry
-
             # Should the boolean be set to False? Does it matter for our purposes?
             states = [(sqlalchemy.inspect(parent), True) for parent in parents]
 
@@ -276,7 +272,7 @@ def _get_relationship_resolver(obj_type, relationship_prop, model_attr):
 
             loader._load_for_path(
                 query_context,
-                path,
+                parent_mapper._path_registry,
                 states,
                 None,
                 child_mapper,
--- a/graphene_sqlalchemy/types.py
+++ b/graphene_sqlalchemy/types.py
@@ -4,10 +4,9 @@ import sqlalchemy
 from promise import dataloader, promise
 from sqlalchemy.ext.hybrid import hybrid_property
 from sqlalchemy.orm import (ColumnProperty, CompositeProperty,
-                            RelationshipProperty, Session)
+                            RelationshipProperty, Session, strategies)
 from sqlalchemy.orm.exc import NoResultFound
 from sqlalchemy.orm.query import QueryContext
-from sqlalchemy.orm.strategies import SelectInLoader
 
 from graphene import Field
 from graphene.relay import Connection, Node
@@ -213,6 +212,8 @@ def _get_custom_resolver(obj_type, orm_field_name):
 def _get_relationship_resolver(obj_type, relationship_prop, model_attr):
     """
     Batch SQL queries using Dataloader to avoid the N+1 problem.
+    SQL batching only works for SQLAlchemy 1.2+ since it depends on
+    the `selectin` loader.
 
     :param SQLAlchemyObjectType obj_type:
     :param sqlalchemy.orm.properties.RelationshipProperty relationship_prop:
@@ -222,7 +223,7 @@ def _get_relationship_resolver(obj_type, relationship_prop, model_attr):
     child_mapper = relationship_prop.mapper
     parent_mapper = relationship_prop.parent
 
-    if relationship_prop.uselist:
+    if not getattr(strategies, 'SelectInLoader', None) or relationship_prop.uselist:
         # TODO Batch many-to-many and one-to-many relationships
         return _get_attr_resolver(obj_type, model_attr, model_attr)
 
@@ -239,7 +240,7 @@ def _get_relationship_resolver(obj_type, relationship_prop, model_attr):
             than re-implementing and maintainnig a big chunk of the `selectin`
             loader logic ourselves.
 
-            The approach is to here to build a regular query that
+            The approach here is to build a regular query that
             selects the parent and `selectin` load the relationship.
             But instead of having the query emits 2 `SELECT` statements
             when callling `all()`, we skip the first `SELECT` statement
@@ -247,6 +248,10 @@ def _get_relationship_resolver(obj_type, relationship_prop, model_attr):
             To accomplish this, we have to construct objects that are
             normally built in the first part of the query in order
             to call directly `SelectInLoader._load_for_path`.
+
+            TODO Move this logic to a util in the SQLAlchemy repo as per
+              SQLAlchemy's main maitainer suggestion.
+              See https://git.io/JewQ7
             """
             session = Session.object_session(parents[0])
 
@@ -258,7 +263,7 @@ def _get_relationship_resolver(obj_type, relationship_prop, model_attr):
                 # The behavior of `selectin` is undefined if the parent is dirty
                 assert parent not in session.dirty
 
-            loader = SelectInLoader(relationship_prop, (('lazy', 'selectin'),))
+            loader = strategies.SelectInLoader(relationship_prop, (('lazy', 'selectin'),))
 
             # Should the boolean be set to False? Does it matter for our purposes?
             states = [(sqlalchemy.inspect(parent), True) for parent in parents]
--- a/graphene_sqlalchemy/types.py
+++ b/graphene_sqlalchemy/types.py
@@ -3,11 +3,12 @@ from collections import OrderedDict
 import sqlalchemy
 from promise import dataloader, promise
 from sqlalchemy.ext.hybrid import hybrid_property
-from sqlalchemy.inspection import inspect as sqlalchemyinspect
-from sqlalchemy.orm import (ColumnProperty, CompositeProperty, Load,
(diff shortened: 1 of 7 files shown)


graphene_sqlalchemy/types.py
```diff
@@ -177,22 +197,132 @@ def construct_fields(
...
+    parent_mapper = relationship_prop.parent
+
+    if relationship_prop.uselist:
+        # TODO Batch many-to-many and one-to-many relationships
+        return _get_attr_resolver(obj_type, model_attr, model_attr)
+
+    class NonListRelationshipLoader(dataloader.DataLoader):
+        cache = False
+
+        def batch_load_fn(self, parents):  # pylint: disable=method-hidden
```
zzzeek: IIUC the purpose of this code is to get to the SelectInLoader._load_for_path method alone; the rest of the private API access here is to try to call into that function in the way that it is usually called, but it appears that this ends up loading in lots of private SQLAlchemy geometries that aren't guaranteed to stay the same.  it is much eaiser for me to provide a public API version of a single call like SelectnLoader._load_for_path (with a fixed single-token path as is the case here) than it is for me to worry that the overall structure of objects like PostLoad / post_load_paths / PostLoad.loaders etc., which are actually all fairly recent additions to the internals, are going to impact external projects if I change them around.     so below i will try to remark on private APIs that don't seem to be needed based on the inputs to this function.

graphene_sqlalchemy/types.py
```diff
@@ -177,22 +197,132 @@ def construct_fields(
...
+            # These issues are very unlikely to happen in practice...
+            for parent in parents:
+                assert parent.__mapper__ is parent_mapper
+                # All instances must share the same session
+                assert session is Session.object_session(parent)
+                # The behavior of `selectin` is undefined if the parent is dirty
+                assert parent not in session.dirty
+
+            load = Load(parent_mapper.entity).selectinload(model_attr)
+            query = session.query(parent_mapper.entity).options(load)
```
zzzeek: so we are starting with a brand new Query here and it has no state or joins at all.  That is, there's no paths or options coming in.   this means a lot of what happens below is much simpler.

graphene_sqlalchemy/types.py
```diff
@@ -177,22 +197,132 @@ def construct_fields(
...
+
+            # Taken from orm.query.Query.__iter__
+            # https://git.io/JeuBi
+            context = query._compile_context()
```
zzzeek: here we are making a QueryContext because it eventually gets passed into _load_for_path, where it is used to determine from the original query what the other MapperOptions on that query are, which above we know they are none, if the Query has _populated_existing on it, which we know it does not, what the current Session is, which we have right here, and that seems to be about it.    So we don't actually need a fully compiled context, we just need an object with basically the Session on it.   You could make a QueryContext directly, or just a short object that has the attributes that the loader needs.    that is still making some private API assumptions, but it would be local to just this one loader method and if i provide a public function it wouldn't need these things.

graphene_sqlalchemy/types.py
```diff
@@ -177,22 +197,132 @@ def construct_fields(
...
+
+            # Taken from orm.loading.instances
+            # https://git.io/JeuBR
+            context.post_load_paths = {}
```
zzzeek: used by PostLoader but if you weren't using PostLoader you wouldn't need to know about this.

graphene_sqlalchemy/types.py
```diff
@@ -177,22 +197,132 @@ def construct_fields(
...
+
+            # Taken from orm.strategies.SelectInLoader.__init__
+            # https://git.io/JeuBd
+            selectin_strategy = getattr(parent_mapper.entity, model_attr).property._get_strategy(load.strategy)
```
zzzeek: you can more directly just instantiate a SelectInLoader object, which means some assumptions about the constructor, but fewer than are needed than calling _get_strategy(load.strategy).

graphene_sqlalchemy/types.py
```diff
@@ -177,22 +197,132 @@ def construct_fields(
...
+
+            # Taken from orm.loading._instance_processor._instance
+            # https://git.io/JeuBq
+            post_load = PostLoad()
```
zzzeek: PostLoad is likely where I'm most uncomfortable with the private API stuff.   this is very esoteric internals that look as strange as they do because they are trying to do as much work up front as is possible, waiting for thousands of rows to come in where we want to minimize the call overhead, so they are hard to follow and they change all the time.     Going straight to _load_for_path, if possible, likely less maintenance, even though that function could change too.

graphene_sqlalchemy/types.py
```diff
@@ -177,22 +197,132 @@ def construct_fields(
...
+                (child_mapper,),
+                {},
+            )
+
+            # Taken from orm.loading._instance_processor._instance
+            # https://git.io/JeuBn
+            # https://git.io/Jeu4j
+            context.partials = {}
+            for parent in parents:
+                post_load.add_state(parent._sa_instance_state, True)
```
zzzeek: _sa_instance_state is accessed publicly by using sqlalchemy.inspect(parent).

graphene_sqlalchemy/types.py
```diff
@@ -177,22 +197,132 @@ def construct_fields(
...
+                parent_mapper,
+                selectin_strategy._load_for_path,
+                (child_mapper,),
+                {},
+            )
+
+            # Taken from orm.loading._instance_processor._instance
+            # https://git.io/JeuBn
+            # https://git.io/Jeu4j
+            context.partials = {}
```
zzzeek: more scary internal stuff that I don't look at unless I have to, I'm not sure if one of these functions hit upon this, I'm not seeing the codepath that does, but this is a longer codepath than I think you need here.

graphene_sqlalchemy/types.py
```diff
@@ -177,22 +197,132 @@ def construct_fields(
...
+            for parent in parents:
+                post_load.add_state(parent._sa_instance_state, True)
+
+            # Taken from orm.strategies.SelectInLoader.create_row_processor
+            # https://git.io/Jeu4F
+            selectin_path = context.query._current_path + parent_mapper._path_registry
```
zzzeek: context.query._current_path is always going to be the "root" path since you created the Query above with no "path", this is PathRegistry.root, but then you are adding parent_mapper._path_registry, but this whole expression can be shortened to parent_mapper._path_registry.

graphene_sqlalchemy/types.py
```diff
@@ -177,22 +198,110 @@ def construct_fields(
...
+            # The path is a fixed single token in this case
+            path = PathRegistry.root + parent_mapper._path_registry
+
+            # Should the boolean be set to False? Does it matter for our purposes?
+            states = [(sqlalchemy.inspect(parent), True) for parent in parents]
+
+            # For our purposes, the query_context will only used to get the session
+            query_context = QueryContext(session.query(parent_mapper.entity))
+
+            loader._load_for_path(
```
zzzeek: great, so you still have some private API here but it's way more contained and we can always make a public function out of this, feel free to propose to SQLAlchemy.

graphene_sqlalchemy/types.py
```diff
@@ -177,22 +197,132 @@ def construct_fields(
...
+    :rtype: Callable
+    """
+    child_mapper = relationship_prop.mapper
+    parent_mapper = relationship_prop.parent
+
+    if relationship_prop.uselist:
+        # TODO Batch many-to-many and one-to-many relationships
+        return _get_attr_resolver(obj_type, model_attr, model_attr)
+
+    class NonListRelationshipLoader(dataloader.DataLoader):
```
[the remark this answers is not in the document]
jnak: Because each relationship is different. The `selectin` loader expects a specific relationship (vs a specific child and / or parent).

jnak:
@Nabellaleen Great that you're on board with the approach. I fixed the tests by disabling batching for SQLAlchemy versions < 1.2. Can you re-approve the diff so I can merge it?

I'm going to send a separate PR soon for batching `one-to-many` and `many-to-many` relationships.

cglacet:
Will this be in the next release?

jnak:
@cglacet Most likely yes. But we're currently blocked by https://github.com/graphql-python/graphene-sqlalchemy/pull/254 and few other dependent changes. Would you mind taking a look at it? That would be really helpful to push this over the finish line

PR #254, opened 2019-11-19
Fix N+1 problem for one-to-many and many-to-many relationships

jnak:
This re-uses all the batching logic implemented in https://github.com/graphql-python/graphene-sqlalchemy/pull/253.

This optimization batches what used to be multiple SQL statements into a single SQL statement. While this this an improvement over the previous behavior, there is still some room for more optimizations: 
- We currently fetch all the children records of the relationship (vs just the ones being requested). Unfortunately, it's not possible to batch pagination with most SQL backends (see Join-Monster [overview of the problem](https://join-monster.readthedocs.io/en/latest/pagination/)). This can gets quite complex so I think we should punt on that for now.
- We currently fetch all the fields of the associated table (vs just the ones being queried). This should be relatively straightforward to implement (though there are some interesting edge cases) so I'm planning to address this soon.

For now, I decided to put this into a separate `BatchSQLAlchemyConnectionField` because the `...ConnectionField` classes need to be refactored a bit in order to allow people to easily opt-out of the batching behavior if need be (via the `ORMField` mechanism). I'll be addressing this in a follow up PR soon. For now, you'll have to enable the optimization via the `SQLAlchemyObjectType.Meta.connection_field_factory` (see `test_batching.py`). 

Cheers

### DIFF
--- a/graphene_sqlalchemy/tests/test_batching.py
+++ b/graphene_sqlalchemy/tests/test_batching.py
@@ -5,9 +5,11 @@ import pkg_resources
 import pytest
 
 import graphene
+from graphene import relay
 
+from ..fields import BatchSQLAlchemyConnectionField
 from ..types import SQLAlchemyObjectType
-from .models import Article, Reporter
+from .models import Article, HairKind, Pet, Reporter
 from .utils import to_std_dicts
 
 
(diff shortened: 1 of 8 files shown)


Nabellaleen:
The approach with using `BatchSQLAlchemyConnectionField` while the problems are not resolved, to allow a opt-in by the implementer is a good idea !

graphene_sqlalchemy/tests/test_batching.py
```diff
@@ -91,8 +81,28 @@ def is_sqlalchemy_version_less_than(version_string):
 
 def test_many_to_one(session_factory):
     session = session_factory()
-    make_fixture(session)
-    schema = get_schema(session)
+
+    reporter_1 = Reporter(
```
Nabellaleen: Why did you "broke" the factorization of this setup code ?

graphene_sqlalchemy/tests/test_batching.py
```diff
@@ -226,3 +256,257 @@ def test_one_to_one(session_factory):
...
+                }
+              }
+            }
+          }
+        """, context_value={"session": session})
+        messages = sqlalchemy_logging_handler.messages
+
+    assert len(messages) == 5
+
+    if is_sqlalchemy_version_less_than('1.3'):
```
Nabellaleen: Should we not also test the SQL generated in 1.2 ?

graphene_sqlalchemy/tests/test_batching.py
jnak: Because different tests needed different fixtures. It's easier this way to keep track of the state of a given test or to add a new test.

graphene_sqlalchemy/tests/test_batching.py
jnak: Generally speaking, I think we should minimize this type of assertions because they are going to be a little annoying to maintain down the road. If anyone updates the base models (e.g. add another column), these tests will have to be updated as well because the generated SQL will be different. 

For `1.3`, it makes sense to explicitly test the SQL because it makes it clear that 5 SQL messages in this context corresponds to 2 `SELECT` statements, that is batching is enabled. That's why it's enough for `1.2` to only assert there are 5 messages because we know it means that batching is enabled. There would be 7 messages (ie 3 SQL statements) without batching.

jnak:
@Nabellaleen Thanks for reviewing. Apologies for the not replying earlier - I got sidetracked by a [concurrency bug](https://github.com/syrusakbary/promise/pull/81) in Promise and Dataloader that was breaking batching.

Once this PR is merged, I'm going to work on setting up benchmarks to compare with and without batching.

graphene_sqlalchemy/tests/test_batching.py
jnak: To workaround the problem mentioned in my previous comment, I added some SQL assertions that should be easy to maintain.

jnak:
I have 2 followup PRs that are currently blocked on this one:
- Add benchmark tests https://github.com/jnak/graphene-sqlalchemy/pull/1
- Add `batching` flag to `SQLAlchemyObjectType` and `ORMField` https://github.com/jnak/graphene-sqlalchemy/pull/2

I'll be sending those once this one is approved because AFAIK Github does not allow working with "stacked" PRs. 

Cheers

PR #259, opened 2020-01-22
Add benchmark for connection fields

jnak:
Add `pytest-benchmark` so we can easily track performance changes over time. 

PS: It's been already approved by @Nabellaleen in my own repo https://github.com/jnak/graphene-sqlalchemy/pull/1

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,9 +5,6 @@ matrix:
     - env: TOXENV=py27
       python: 2.7
     # Python 3.5
-    - env: TOXENV=py34
-      python: 3.4
-    # Python 3.5
     - env: TOXENV=py35
       python: 3.5
     # Python 3.6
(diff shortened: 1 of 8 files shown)


jnak:
@Nabellaleen For some reasons that I don't fully understand, our tests for Python 3.4 suddenly stopped working [on master](https://travis-ci.org/graphql-python/graphene-sqlalchemy/builds/640676996). It can no longer install `PyYaml`, a dependency of `pre-commit`, one our dev dependencies. I investigated a bit and it looks like it's due to an [underlying issue](https://github.com/yaml/pyyaml/issues/281) in PyYaml. Given Python 3.4 has reached end of life [in March 2019](https://devguide.python.org/#status-of-python-branches) and it's fairly easy to upgrade from Python 3.4, I don't think we should spend time trying to make the tests pass. I disabled the tests for Python 3.4. Nonetheless, I still expect `graphene-sqlalchemy` to work with Python 3.4 for the foreseeable future given that we don't use anything specific to Python 3.4.

PR #260, opened 2020-01-24
Add `batching` params 

jnak:
Add parameters to toggle `batching` on or off. This can be configured at 2 levels:
- we can configure all the fields of a type at once via `SQLAlchemyObjectType.meta.batching` 
- or we can specify it for a specific field via `ORMfield.batching`. This trumps `SQLAlchemyObjectType.meta.batching`.

### DIFF
--- a/graphene_sqlalchemy/converter.py
+++ b/graphene_sqlalchemy/converter.py
@@ -34,47 +34,79 @@ def is_column_nullable(column):
 
 
 def convert_sqlalchemy_relationship(relationship_prop, obj_type, connection_field_factory, batching,
-                                    attr_name, orm_field_name, **field_kwargs):
+                                    orm_field_name, **field_kwargs):
     """
     :param sqlalchemy.RelationshipProperty relationship_prop:
-    :param Registry registry:
-    :type function|None connection_field_factory:
-    :type bool batching:
     :param SQLAlchemyObjectType obj_type:
+    :param function|None connection_field_factory:
+    :param bool batching:
     :param str orm_field_name:
+    :param dict field_kwargs:
     :rtype: Dynamic
     """
     def dynamic_type():
+        """:rtype: Field|None"""
         direction = relationship_prop.direction
-        model = relationship_prop.mapper.entity
-        type_ = obj_type._meta.registry.get_type_for_model(model)
-
+        child_type = obj_type._meta.registry.get_type_for_model(relationship_prop.mapper.entity)
         batching_ = batching if is_selectin_available else False
-        connection_field_factory_ = connection_field_factory
 
-        if not type_:
+        if not child_type:
             return None
 
         if direction == interfaces.MANYTOONE or not relationship_prop.uselist:
-            resolver = get_custom_resolver(obj_type, orm_field_name)
-            if resolver is None:
-                resolver = get_batch_resolver(relationship_prop) if batching_ else \
-                    get_attr_resolver(obj_type, relationship_prop.key)
-
-            return Field(type_, resolver=resolver, **field_kwargs)
+            return _convert_o2o_or_m2o_relationship(relationship_prop, obj_type, batching_, orm_field_name,
+                                                    **field_kwargs)
 
         if direction in (interfaces.ONETOMANY, interfaces.MANYTOMANY):
-            if not type_._meta.connection:
-                return Field(List(type_), **field_kwargs)
-
-            if connection_field_factory_ is None:
-                connection_field_factory_ = BatchSQLAlchemyConnectionField.from_relationship if batching_ else \
-                    default_connection_field_factory
-
-            # TODO Allow override of connection_field_factory and resolver via ORMField
-            return connection_field_factory_(relationship_prop, obj_type._meta.registry, **field_kwargs)
+            return _convert_o2m_or_m2m_relationship(relationship_prop, obj_type, batching_,
+                                                    connection_field_factory, **field_kwargs)
 
     return Dynamic(dynamic_type)
+
+
+def _convert_o2o_or_m2o_relationship(relationship_prop, obj_type, batching, orm_field_name, **field_kwargs):
+    """
+    Convert one-to-one or many-to-one relationshsip. Return an object field.
+
+    :param sqlalchemy.RelationshipProperty relationship_prop:
+    :param SQLAlchemyObjectType obj_type:
+    :param bool batching:
+    :param str orm_field_name:
+    :param dict field_kwargs:
+    :rtype: Field
+    """
+    child_type = obj_type._meta.registry.get_type_for_model(relationship_prop.mapper.entity)
+
+    resolver = get_custom_resolver(obj_type, orm_field_name)
+    if resolver is None:
+        resolver = get_batch_resolver(relationship_prop) if batching else \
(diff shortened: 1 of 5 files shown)


graphene_sqlalchemy/converter.py
```diff
@@ -26,29 +33,46 @@ def is_column_nullable(column):
...
+    """
     def dynamic_type():
-        _type = registry.get_type_for_model(model)
+        direction = relationship_prop.direction
+        model = relationship_prop.mapper.entity
+        type_ = obj_type._meta.registry.get_type_for_model(model)
 
-        if not _type:
+        batching_ = batching if is_selectin_available else False
+        connection_field_factory_ = connection_field_factory
```
Nabellaleen: Why do you need this allocation ?

graphene_sqlalchemy/converter.py
```diff
@@ -26,29 +33,46 @@ def is_column_nullable(column):
...
-        elif direction in (interfaces.ONETOMANY, interfaces.MANYTOMANY):
-            if _type._meta.connection:
-                # TODO Add a way to override connection_field_factory
-                return connection_field_factory(relationship_prop, registry, **field_kwargs)
-            return Field(
-                List(_type),
-                **field_kwargs
-            )
+            resolver = get_custom_resolver(obj_type, orm_field_name)
+            if resolver is None:
```
Nabellaleen: I don't find this syntax very readable. maybe something like that ?
```suggestion
            if resolver is None and batching_: 
```

graphene_sqlalchemy/converter.py
```diff
@@ -26,29 +33,46 @@ def is_column_nullable(column):
...
+                resolver = get_batch_resolver(relationship_prop) if batching_ else \
```
Nabellaleen:
```suggestion
                resolver = get_batch_resolver(relationship_prop)
```

graphene_sqlalchemy/converter.py
```diff
@@ -26,29 +33,46 @@ def is_column_nullable(column):
...
+                    get_attr_resolver(obj_type, relationship_prop.key)
```
Nabellaleen:
```suggestion
            elif resolver is None:
                resolver = get_attr_resolver(obj_type, relationship_prop.key) 
```

graphene_sqlalchemy/converter.py
```diff
@@ -26,29 +33,46 @@ def is_column_nullable(column):
...
+
+            return Field(type_, resolver=resolver, **field_kwargs)
+
+        if direction in (interfaces.ONETOMANY, interfaces.MANYTOMANY):
+            if not type_._meta.connection:
+                return Field(List(type_), **field_kwargs)
+
+            if connection_field_factory_ is None:
+                connection_field_factory_ = BatchSQLAlchemyConnectionField.from_relationship if batching_ else \
+                    default_connection_field_factory
```
Nabellaleen: And same thing here, I think :)

graphene_sqlalchemy/converter.py
jnak: This is because I need to modify `connection_field_factory` in some cases (see [below](https://github.com/graphql-python/graphene-sqlalchemy/pull/260/files#r370926773)) and I can't reassign to `connection_field_factory` since it is defined outside of the scope `dynamic`. Otherwise, you get: `UnboundLocalError: local variable 'connection_field_factory' referenced before assignment`.

graphene_sqlalchemy/converter.py
jnak: That would be a different behavior. To get the same behavior, I would need to do something like that:
```py
if resolver is None and batching_:
  resolver = get_batch_resolver(relationship_prop)
elif resolver is None and not batching_:
   resolver = get_attr_resolver(obj_type, relationship_prop.key)
```
I think I prefer the current version because it reads to me more like the intended logic:
> If no custom resolver was defined by the user, 
> then set it to batch resolver if batching is enabled, otherwise set it to the regular resolver

That said, I don't feel strongly about it so let me know what you prefer.

graphene_sqlalchemy/converter.py
jnak: See my comment above

jnak:
@Nabellaleen Ignore @katy-sadowski approval - I asked her to do this so I can tests different permission settings.

graphene_sqlalchemy/converter.py
Nabellaleen: Yeah, I understand your point, indeed
The main problem, finally, is the "complexity" due to multiple indentation, which could resolve by a more "splited" code.

Maybe not in this PR, but this block could be splited in smaller chunks.

jnak: I see. I tried to extract some logic into sub-functions it but I'm not sure I like this because those sub-functions don't seem to make much sense on their own: 
```py
def convert_sqlalchemy_relationship(relationship_prop, obj_type, connection_field_factory, batching,
                                    attr_name, orm_field_name, **field_kwargs):
    def dynamic_type():
        direction = relationship_prop.direction
        model = relationship_prop.mapper.entity
        type_ = obj_type._meta.registry.get_type_for_model(model)

        batching_ = batching if is_selectin_available else False
        connection_field_factory_ = connection_field_factory

        if not type_:
            return None

        if direction == interfaces.MANYTOONE or not relationship_prop.uselist:
            resolver = _get_resolver_for_o2o_or_m2o_relationship(obj_type, orm_field_name, relationship_prop, batching_)
            return Field(type_, resolver=resolver, **field_kwargs)

        if direction in (interfaces.ONETOMANY, interfaces.MANYTOMANY):
            if not type_._meta.connection:
                return Field(List(type_), **field_kwargs)

            connection_field_factory_ = _get_connection_field_factory(connection_field_factory_, batching_)
            return connection_field_factory_(relationship_prop, obj_type._meta.registry, **field_kwargs)

    return Dynamic(dynamic_type)


def _get_resolver_for_o2o_or_m2o_relationship(obj_type, orm_field_name, relationship_prop, batching):
    resolver = get_custom_resolver(obj_type, orm_field_name)
    if resolver is None:
        return get_batch_resolver(relationship_prop) if batching else \
            get_attr_resolver(obj_type, relationship_prop.key)


def _get_connection_field_factory(connection_field_factory, batching):
    # TODO Allow override of connection_field_factory and resolver via ORMField
    if connection_field_factory is None:
        connection_field_factory = BatchSQLAlchemyConnectionField.from_relationship if batching else \
            default_connection_field_factory
    
    return connection_field_factory
```
You really need to understand the logic of `convert_sqlalchemy_relationship` to understand `_get_resolver_for_o2o_or_m2o_relationship` and `_get_connection_field_factory`. That said, it makes `convert_sqlalchemy_relationship` reading a bit easier in the sense that you can clearly see the 3 main `if` statements. 

What are you thoughts? I could go either way on this.

Nabellaleen: Yeah, indeed these 2 methods are not really understandable on their own, but as it's spliting the code and allowing a easier reading, I think it could be a good "first step" for a big improvement : the next time someone will modify this code, it will be easier to go a step further

jnak: Fair point. I ended up going a step further.

himat:
Was the default batching true or false?

PR #267, opened 2020-02-12
association_proxy support

### LINKED ISSUE #136: Support for associationproxy
Hi,

I'm using graphene-sqlalchemy 2.0.0. I haven't found any documentation around association proxy support and my preliminary efforts into finding it in the code/docs/issues have proved futile. Are association proxies supported? If not, is there a plan to support them?

Thanks!

dpep:
support for sqlalchemy AssociationProxy properties (https://docs.sqlalchemy.org/en/13/orm/extensions/associationproxy.html)

Fixes #136


dpep:
@jnak : any initial thoughts on adding AssociationProxy support?

erikwrede:
Just checked your PR out. It's a feature I could use as well! I know it's been some time, but since checks are passing, I'm going to test this a bit in the upcoming days and give you a review. Confident we can get this merged soon.

erikwrede (changes requested):
It took me longer than expected, but the rest looks fine for basic feature integration.  We should add Hybrid Property support before this gets merged.
Thanks for the great work; this is helpful! If you @dpep have the time to implement the additional features, I'd highly appreciate that. Otherwise, I will work on it myself once I find the time.

erikwrede:
Quick update: using this in dev, but there seem to be problems regarding batching/dataloader, especially if the association proxy "skips" an intermediate class in a relationship and maps to another relationship (A.b->B.c->C => A.c->C). Will need to investigate further once I find the time.

erikwrede:
Fixed the review comment I've had. The issue with dataloaders should be solved in a follow-up PR. We could easily create a optional custom resolver for the proxies that first calls the dataloader for the specific relationship, so no implicit load takes place upon access.

jendrikjoe:
Generally like the change as it probably saves a lot of code. One thing we have to be cautious about is that when variables are proxied they might be proxied between nodes. This will eventually conflict with the caching relay and apollo do internally. Imagine the following classes:
```python
class A:
    id = str()
    b = relationship(B)
    proxied_c = association_proxy(B.c)
class B:
    id = str()
    c = str()
```

If we now run the mutation:
```
    query = """mutation alterB($c: str!) {
            alterB(c: $c) {
                id
                c
            }
        }"""
```
the cache for B is update to contain the new value of c but the cache for a, which might have been queried before wouldn’t be updated. We just have to make sure to document this possible conflict, as it probably goes a bit against the patter of just having on root per property. What do you think? 🙂

erikwrede:
Good idea to mention that in the docs, probably best to scope that to an additional PR. It's more of a frontend task, but we cannot expect all of our users to know all about the best practices and patterns in GQL upfront. Let's address this after the release of 3.0 @jendrikjoe 👍

jendrikjoe:
Sounds reasonable 👍 The rest of the code is fine by me 👍

conao3:
I now find `association_proxy` and I want to use it with `graphene-sqlalchemy`.
What status on this PR? There're conflict, we cannot merge this?

erikwrede:
@conao3 It's on my list to get it done for this month. If you have time to test it, I'd really appreciate a review! ☺️

conao3:
This patch is already cherrypicked own fork and it's ready to ship :)
This seems super good but one point, I have to address `required` argument but I think we can fix it as a single issue after the merge.

erikwrede:
@conao3 how would you addrss the `required` argument?

conao3:
I want to use this patch now, I haven't investigated it.

The issue I face is, the `association_proxy` column is always `required = True`, but my `association_proxy` was a proxy for `relationship_ship`, so it had to be `required=False`. I decided to always specify `required=False` for the time being to prevent graphene from generating an error.

erikwrede:
@conao3 I just checked, and I can't reproduce that behavior. For me, association proxies are always `Nullable`. That issue might be related to you local configuration.

conao3:
OK, thanks for letting me know!

erikwrede (approved):

PR #299, opened 2021-02-04
Simplify import of version number

dotcs:
I found the previous implementation to import the version number too complex, but maybe I have missed something here? If possible this should be simplified.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -1,15 +1,5 @@
-import ast
-import re
-import sys
-
 from setuptools import find_packages, setup
-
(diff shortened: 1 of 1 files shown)


bim9262:
The reason that's it's written the way it is is that graphene_sqlalchemy/__init__.py imports things that will be installed by setup.py, but aren't installed at the time that you tried to do the import.

dotcs:
Alright, thanks for the clarification. I'll close this PR then.

PR #301, opened 2021-03-19
Enum name for sort fields should be taken from Column.key instead of Column.name

davidcim:
I'm working with an external Microsoft SQL database and some field names contain invalid characters such as `%`. Fortunately, the parameter `key` of SQLAlchemy [Column](https://docs.sqlalchemy.org/en/13/core/metadata.html?highlight=column#sqlalchemy.schema.Column) allows you to specify a different name for a field to be used everywhere except in the SQL code:

>> **key** – An optional string identifier which will identify this Column object on the Table. When a key is provided, this is the only identifier referencing the Column within the application, including ORM attribute mapping; the name field is used only when rendering SQL.

Usually the `key` attribute in the orm column is the same as the `name` attribute. But, as in this instance, if you specify the key parameter when defining the column:
```python
descuento1 = Column("% descuento1", Numeric, key="descuento1")
```
... `sort_enum_for_object_type` function will generate the name of the sort enum from the original SQL field name instead of from the provided `key` alternative as it would be expected. As a result, the generated enum name is wrong and an exception is thrown:
```
AssertionError: Names must match /^[_a-zA-Z][_a-zA-Z0-9]*$/ but "%_DESCUENTO1_ASC" does not.
```
This PR just makes the name to be taken from the `key` attribute which is usually identical except if an alternative field name is specified. Everything seems to work fine in my tests and I can't find any side effects resulting from the change.

`tox -e pre-commit`:
```
Check for merge conflicts................................................Passed
Check Yaml...............................................................Passed
Debug Statements (Python)................................................Passed
Fix End of Files.........................................................Passed
Trim Trailing Whitespace.................................................Passed
flake8...................................................................Passed
seed isort known_third_party.............................................Passed
isort....................................................................Passed
```


erikwrede:
Thanks for the PR; I just checked it out!
I agree with your points. The current behavior is intransparent because it can cause sort enum names to differ from the field's names in the schema.
However, since the rest of the library uses the field names instead of the column names, wouldn't it make more sense to use the field name here instead of the `column.key` to handle naming uniformly everywhere?
Functionally, I don't think it makes a difference since `column.key` seems to be initialized to the field name instead of `column.name` if no custom key is provided.
It's just a matter of handling names the same way in the entire codebase to prevent future deviations like this. 
I must admit, it's a bit of nitpicking, but I'd be happy to have your opinion on this.

I've also found another case where `column.name` is used: 
https://github.com/graphql-python/graphene-sqlalchemy/blob/a47dbb369c5e7a057affe7cb98fa15d3acf29cd8/graphene_sqlalchemy/converter.py#L241-L249

It looks like that could cause the same errors, so it might be worth fixing as well.

davidcim:
Erik,
Sorry I didn't reply sooner, I'm swamped with work. If I have time I will make a new PR with the changes you suggest and tests.
Regards.

erikwrede:
Commit in PR is: https://github.com/graphql-python/graphene-sqlalchemy/pull/353/commits/dccbe227b1e5a0118061de5840dde81154771e5a

Issue #329, opened 2022-01-27
BUG:  Out of Sync with graphql-relay  > 3.1


ed-turner:
This is related to the import `graphql_relay.connection.arrayconnection` in `graphene_sqlalchemy.fields.SQLAlchemyConnectionField`.  

To reproduce:

1. pip install graphql-relay==3.1.0
2. Run `from graphene_sqlalchemy.fields import SQLAlchemyConnectionField`
3. You will get an import module error for `graphql_relay.connection.arrayconnection` 

One suggestion is to either make specific requirements for graphql-relay in the install requirements or update the imports in the 3.0 release.

Cito:
There will be a patch release of graphql-relay 3.1 for backward compatibility today, but it should also be fixed here by importing from the top level of graphql-relay.

Cito:
Fix available in PR #330

PR #330, opened 2022-01-28
Fix for import from graphql-relay-py (#329)

Cito:


conao3:
I want to this patch for my production source code to reduce unneeded warning. Any blockers?

erikwrede (approved):

PR #333, opened 2022-03-04
Get default type of hybrid_property from type annotation

conao3:
Fix below TODO comment.
```
# TODO The default type should be dependent on the type of the property propety.
```

Examines the return type of the function of `hybrid_property` and converts it to the default graphene type.

### DIFF
--- a/graphene_sqlalchemy/converter.py
+++ b/graphene_sqlalchemy/converter.py
@@ -260,27 +260,27 @@ def convert_json_type_to_string(type, column, registry=None):
     return JSONString
 
 
-def convert_hybrid_type(type, column):
-    if type in [str, datetime.date, datetime.time]:
+def convert_hybrid_type(type_, column):
+    if type_ in [str, datetime.date, datetime.time]:
         return String
-    elif type == datetime.datetime:
+    elif type_ == datetime.datetime:
         from graphene.types.datetime import DateTime
         return DateTime
-    elif type == int:
+    elif type_ == int:
         return Int
-    elif type == float:
+    elif type_ == float:
         return Float
-    elif type == bool:
+    elif type_ == bool:
         return Boolean
-    elif getattr(type, "__origin__", None) == list:  # check for typing.List[T]
-        args = getattr(type, "__args__", [])
+    elif getattr(type_, "__origin__", None) == list:  # check for typing.List[T]
+        args = getattr(type_, "__args__", [])
         if len(args) != 1:
             return String  # Unknown fallback
 
         inner_type = convert_hybrid_type(args[0], column)
         return List(inner_type)
-    elif getattr(type, "__name__", None) == "list":  # check for list[T]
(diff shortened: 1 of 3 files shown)


conao3:
```
In modern python, instead of a class, it contains class name
```
This commit message is wrong. The truth is, 
```
Under `from __future__ import annotations`, instead of a class, it contains class name
```

erikwrede:
I just looked at your PR, I think it's a useful change!

Is there any reason you're selecting the corresponding Graphene Types using a switch/case statement instead of adapting the annotation-based system that is used to register SQLAlchemy types? (`@convert_sqlalchemy_type.register(JSONType)`)

I think creating an annotation for the hybrid properties (e.g. something like `@convert_sqlalchemy_type.register_hybrid(Type)` ) would allow for more flexibility, and is IMO the preferred option over statically converting the base types, since it would allow users to add their own support for types like Tuples etc until we include them. What do you think?

graphene_sqlalchemy/converter.py
```diff
@@ -256,3 +259,40 @@ def convert_json_to_string(type, column, registry=None):
 @convert_sqlalchemy_type.register(JSONType)
 def convert_json_type_to_string(type, column, registry=None):
     return JSONString
+
+
+def _check_type(type_, args):
+    return type_ in (args + [x.__name__ for x in args])
+
+
+def convert_hybrid_type(type_, column):
```
erikwrede, line 268: See my comment in the PR

flipbit03: I need this feature, so I picked it up to try to finish the work.

Just as a heads up, I tried to use the actual `@singledispatch` mechanism and it is not compatible, since it actually extracts the type of the first argument, so for example, from a function's return type of `str`, this value correctly becomes `type` (being the type of the str class) :cry: 

I'm trying to code an approach similar to that where we can decorate functions and switch based on value.

erikwrede: Let me know what you come up with! 😃

flipbit03: I will, thank you :pray: I'll open a new Pull Request for #333 from my own branch as soon as I have the basics working.

conao3: I want to finish this work. I will work on this PR tomorrow, so can it wait?

conao3: I am branching by case statement because it returning the corresponding Graphene type in a class of `type` type obtained by `__annotations__`.
`functools.singledispatch` is a way to achieve polymorphism by instance type, but I don't know at this time how to achieve polymorphism by the class itself.

flipbit03: I finished the work. I am opening a pull request against this issue.

erikwrede:
Thanks again for initiating this @conao3!

conao3:
Thank you, too!

PR #355, opened 2022-07-23
Enable sorting when batching is enabled

PaulSchweizer:
This is a followup of our discussion from July 18th about batching and sorting. 
I took a deeper dive into the batching feature and have a better understanding of it now.
Basically the core issue was, that the batching resolver did not take care of running the initial query. Instead, the initial query had to be run "manually" in a custom "resolve_" field as indicated in the tests: https://github.com/graphql-python/graphene-sqlalchemy/blob/dfee3e9417cdb8a6ec67b5cd79ee203ce4f72ed7/graphene_sqlalchemy/tests/test_batching.py#L67
That of course bypassed the "get_query" function in the resolver and thus also made sorting "impossible".
Let me know what you think.

### DIFF
--- a/graphene_sqlalchemy/batching.py
+++ b/graphene_sqlalchemy/batching.py
@@ -1,3 +1,5 @@
+from asyncio import get_event_loop
+
 import aiodataloader
 import sqlalchemy
 from sqlalchemy.orm import Session, strategies
@@ -5,15 +7,21 @@ from sqlalchemy.orm.query import QueryContext
 
 from .utils import is_sqlalchemy_version_less_than
 
+# Cache this across `batch_load_fn` calls
+# This is so SQL string generation is cached under-the-hood via `bakery`
+# Caching the relationship loader for each relationship prop.
+RELATIONSHIP_LOADERS_CACHE = {}
+
 
 def get_batch_resolver(relationship_prop):
 
-    # Cache this across `batch_load_fn` calls
-    # This is so SQL string generation is cached under-the-hood via `bakery`
-    selectin_loader = strategies.SelectInLoader(relationship_prop, (('lazy', 'selectin'),))
-
     class RelationshipLoader(aiodataloader.DataLoader):
         cache = False
+
+        def __init__(self, relationship_prop, selectin_loader):
+            super().__init__()
+            self.relationship_prop = relationship_prop
+            self.selectin_loader = selectin_loader
 
         async def batch_load_fn(self, parents):
             """
@@ -38,8 +46,8 @@ def get_batch_resolver(relationship_prop):
               SQLAlchemy's main maitainer suggestion.
               See https://git.io/JewQ7
             """
-            child_mapper = relationship_prop.mapper
-            parent_mapper = relationship_prop.parent
+            child_mapper = self.relationship_prop.mapper
+            parent_mapper = self.relationship_prop.parent
             session = Session.object_session(parents[0])
 
             # These issues are very unlikely to happen in practice...
@@ -62,7 +70,7 @@ def get_batch_resolver(relationship_prop):
                 query_context = parent_mapper_query._compile_context()
 
             if is_sqlalchemy_version_less_than('1.4'):
-                selectin_loader._load_for_path(
+                self.selectin_loader._load_for_path(
                     query_context,
                     parent_mapper._path_registry,
                     states,
@@ -70,7 +78,7 @@ def get_batch_resolver(relationship_prop):
                     child_mapper
                 )
             else:
-                selectin_loader._load_for_path(
+                self.selectin_loader._load_for_path(
                     query_context,
                     parent_mapper._path_registry,
                     states,
@@ -78,10 +86,26 @@ def get_batch_resolver(relationship_prop):
                     child_mapper,
                     None
                 )
+            return [getattr(parent, self.relationship_prop.key) for parent in parents]
 
-            return [getattr(parent, relationship_prop.key) for parent in parents]
+    def _get_loader(relationship_prop):
+        """Retrieve the cached loader of the given relationship."""
+        loader = RELATIONSHIP_LOADERS_CACHE.get(relationship_prop, None)
+        if loader is None:
+            selectin_loader = strategies.SelectInLoader(
+                relationship_prop,
+                (('lazy', 'selectin'),)
+            )
+            loader = RelationshipLoader(
+                relationship_prop=relationship_prop,
+                selectin_loader=selectin_loader
+            )
+            RELATIONSHIP_LOADERS_CACHE[relationship_prop] = loader
+        else:
+            loader.loop = get_event_loop()
+        return loader
 
-    loader = RelationshipLoader()
+    loader = _get_loader(relationship_prop)
 
     async def resolve(root, info, **args):
         return await loader.load(root)
--- a/graphene_sqlalchemy/batching.py
+++ b/graphene_sqlalchemy/batching.py
@@ -1,4 +1,6 @@
+"""The dataloader uses "select in loading" strategy to load related entities."""
 from asyncio import get_event_loop
+from typing import Dict
 
 import aiodataloader
 import sqlalchemy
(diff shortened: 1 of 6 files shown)


graphene_sqlalchemy/batching.py
```diff
@@ -1,20 +1,28 @@
...
 import sqlalchemy
 from sqlalchemy.orm import Session, strategies
 from sqlalchemy.orm.query import QueryContext
 
 from .utils import is_sqlalchemy_version_less_than
 
+# Cache this across `batch_load_fn` calls
+# This is so SQL string generation is cached under-the-hood via `bakery`
+# Caching the relationship loader for each relationship prop.
+RELATIONSHIP_LOADERS_CACHE = {}
```
PaulSchweizer, line 13: We store the loaders per relationship so each item accesses the same cached result ensuring that the selectin is only triggered once

graphene_sqlalchemy/batching.py
```diff
@@ -62,26 +70,42 @@ async def batch_load_fn(self, parents):
...
+                relationship_prop,
+                (('lazy', 'selectin'),)
+            )
+            loader = RelationshipLoader(
+                relationship_prop=relationship_prop,
+                selectin_loader=selectin_loader
+            )
+            RELATIONSHIP_LOADERS_CACHE[relationship_prop] = loader
+        else:
+            loader.loop = get_event_loop()
```
PaulSchweizer, line 105: I'm not 100% sure about this. Do any of you see issues with this?

graphene_sqlalchemy/fields.py
```diff
@@ -129,20 +129,31 @@ def get_query(cls, model, info, sort=None, **args):
         return query
 
 
-class BatchSQLAlchemyConnectionField(UnsortedSQLAlchemyConnectionField):
+class BatchSQLAlchemyConnectionField(SQLAlchemyConnectionField):
```
PaulSchweizer, line 132: I'm not sure what our exact plan in regards to unifying these fields was. I just made this sortable, we can discuss this again

graphene_sqlalchemy/tests/models.py
```diff
@@ -110,6 +110,24 @@ class Article(Base):
...
+class Reader(Base):
+    __tablename__ = "readers"
+    id = Column(Integer(), primary_key=True)
+    name = Column(String(100))
+    articles = relationship(
+        "Article", secondary="articles_readers", back_populates="readers"
+    )
+
+
+class ArticleReader(Base):
```
PaulSchweizer, line 127: Had to add some deeper nesting in the test data for more robust tests

graphene_sqlalchemy/tests/test_batching.py
```diff
@@ -73,6 +73,40 @@ def resolve_reporters(self, info):
...
+
+    class ReaderType(SQLAlchemyObjectType):
+        class Meta:
+            model = Reader
+            name = "Reader"
+            interfaces = (relay.Node,)
+            batching = True
+            connection_class = Connection
+
+    class Query(graphene.ObjectType):
```
PaulSchweizer, line 101: As mentioned, the original test schema defines custom resolve_ methods which I find odd, but maybe people use it like that. We use it like this.

graphene_sqlalchemy/tests/test_batching.py
```diff
@@ -82,11 +116,11 @@ async def test_many_to_one(session_factory):
     session = session_factory()
 
     reporter_1 = Reporter(
-      first_name='Reporter_1',
+        first_name='Reporter_1',
     )
     session.add(reporter_1)
     reporter_2 = Reporter(
-      first_name='Reporter_2',
+        first_name='Reporter_2',
```
PaulSchweizer, line 123: unfortunately there were some flake8 errors I had to fix, so ignore these indentation changes

graphene_sqlalchemy/batching.py
erikwrede: Why do we need the get_event_loop() call in the else part? Isn't that initially covered by `aiodataloader` already, and shouldn't the cache also get a full reset if the event loop changes (implying thread restart/change)?

`aiodataloader` recommends to create a new loader for each request, this would keep the loaders between different queries, right?

graphene_sqlalchemy/fields.py
erikwrede: Looks good.
If I'm correct, this would remove the need for the separate unsorted field as sorting for all the possible fields could be disabled using the `Meta` options. Maybe there is a need to add a QoL `sort=None` shortcut for the  options.

erikwrede:
https://github.com/graphql-python/graphene-sqlalchemy/blob/ac57fd4ba78a786f36aa0c948596a66766c33ec7/graphene_sqlalchemy/enums.py#L147

Changing this to `field_name` would fix the errors with sqa-13.
It seems like `column_prop` of `Reporter` has the name set to something similar to `%(4447102096 anon)s` at the time of sort enum generation. This causes `def sort_enum_for_object_type` to generate a sort enum with field names not allowed by GQL (`%(4447102096 anon)s_ASC  & %(4447102096 anon)s_DESC`), causing the test failure. Couldn't figure out why `get_schema` does not reproduce the same behavior although the same models were used.

graphene_sqlalchemy/batching.py
PaulSchweizer: Makse sense, I changed it

graphene_sqlalchemy/fields.py
PaulSchweizer: I deprecated the UnsortedSQLAlchemyConnectionField as you can just set `sort=None` on the meta class already to disable sorting: https://github.com/graphql-python/graphene-sqlalchemy/blob/6417061998823fb3531def8fee394bd81c6f4c24/graphene_sqlalchemy/tests/test_sort_enums.py#L263

PaulSchweizer:
I still need to investigate those failing tests

PaulSchweizer:
The tests pass now, the remaining issue was that sqla1.2 is creating slighlty different select statments so my asserts where not working there. This is ready from my side now, please give it another look.

erikwrede (changes requested):
Tested this locally. Working nicely!
I suggested a few minor adjustments. Please check my comments.

graphene_sqlalchemy/batching.py
```diff
@@ -6,82 +10,108 @@
...
+            )
+        return [
+            getattr(parent, self.relationship_prop.key) for parent in parents
+        ]
+
+
+# Cache this across `batch_load_fn` calls
+# This is so SQL string generation is cached under-the-hood via `bakery`
+# Caching the relationship loader for each relationship prop.
+RELATIONSHIP_LOADERS_CACHE: Dict[
```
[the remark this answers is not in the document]
PaulSchweizer: The cache is necessary because during a request, each `resolve` needs to access the same loader so it can draw on the same cache. Without keeping a reference to the loader between resolves, we would create a new loader for each `resolve` and thus not be able to draw on the loader's cache.

Note though, that the loader is recreated for each request, so it is not really "cached" between requests, the loader object just lingers around in this dict until the next request. I'm not sure how I could clear this cache after the request has been resolved if that is what you where thinking about. You'd have to give me a pointer where and how that could be implemented.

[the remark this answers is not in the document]
erikwrede: I didn't see that the loader is recreated for each request. I think we can leave it as it is then. The cache not being cleared shouldn't cause any additional trouble.

graphene_sqlalchemy/fields.py
```diff
@@ -90,59 +122,43 @@ def wrap_resolve(self, parent_resolver):
         )
 
 
-# TODO Rename this to SortableSQLAlchemyConnectionField
-class SQLAlchemyConnectionField(UnsortedSQLAlchemyConnectionField):
+# TODO Remove in next major version
+class UnsortedSQLAlchemyConnectionField(SQLAlchemyConnectionField):
```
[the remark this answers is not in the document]
PaulSchweizer: I added your suggestion and a test for it.

graphene_sqlalchemy/enums.py
```diff
@@ -144,9 +144,9 @@ def sort_enum_for_object_type(
             column = orm_field.columns[0]
             if only_indexed and not (column.primary_key or column.index):
                 continue
-            asc_name = get_name(column.key, True)
```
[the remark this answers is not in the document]
sabard: This change looks reasonable to me. The only place I can see there being an issue is if a user defined a custom `ORMField` attribute on the ObjectType or any parent class (see [types.py#L128](https://github.com/graphql-python/graphene-sqlalchemy/blob/515060b6ea37f2fa32855bfd8a4c8c49bf5bc2e8/graphene_sqlalchemy/types.py#L128)) with a different name from the field it references. I can write a test case to see what happens in this case or we can just include a note in the release.

graphene_sqlalchemy/tests/models.py
sabard: I added a `Tags` class in #357. Will see if there's a way to consolidate over there after this is merged.

sabard (commented):
Aside from Erik's comments, all LGTM! Let me know if you'd like me to write a test case for custom `ORMField` attributes.

PaulSchweizer:
It would be great @sabard, if you could write the test for the potential issue you have in mind. If it fails I will look into it and work on a solution.

graphene_sqlalchemy/tests/test_batching.py
```diff
@@ -531,6 +565,70 @@ def resolve_reporters(self, info):
     assert len(select_statements) == 2
 
 
+@pytest.mark.asyncio
+def test_batch_sorting_with_custom_ormfield(session_factory):
```
sabard, line 569: This works as expected as far as I can tell. You can sort with `FIRSTNAME_*` or `FIRST_NAME_*` which is still automatically created.

graphene_sqlalchemy/tests/test_batching.py
```diff
@@ -531,6 +565,70 @@ def resolve_reporters(self, info):
...
+            {"node": {
+                "firstname": "Reporter_2",
+            }},
+            {"node": {
+                "firstname": "Reporter_1",
+            }},
+        ]}
+    }
+    select_statements = [message for message in messages if 'SELECT' in message and 'FROM reporters' in message]
+    assert len(select_statements) == 2
```
sabard, line 629: Do you all have any idea why this results in 2 select statements? Seems to be a relay thing. Full SQL if helpful:

```SQL
SELECT count(*) AS count_1 
FROM (SELECT (SELECT CAST(count(reporters.id) AS INTEGER) AS count_2 
FROM reporters) AS anon_2, reporters.id AS reporters_id, reporters.first_name AS reporters_first_name, reporters.last_name AS reporters_last_name, reporters.email AS reporters_email, reporters.favorite_pet_kind AS reporters_favorite_pet_kind 
FROM reporters ORDER BY reporters.first_name DESC) AS anon_1

SELECT (SELECT CAST(count(reporters.id) AS INTEGER) AS count_1 
FROM reporters) AS anon_1, reporters.id AS reporters_id, reporters.first_name AS reporters_first_name, reporters.last_name AS reporters_last_name, reporters.email AS reporters_email, reporters.favorite_pet_kind AS reporters_favorite_pet_kind 
FROM reporters ORDER BY reporters.first_name DESC
LIMIT ? OFFSET ?

```

sabard (commented):
Looks like all works well for custom `ORMField` names.

Off topic, but is sorting supported on non-Relay-style batching? I wasn't able to get it to work; maybe deserves its own test if supported?

graphene_sqlalchemy/tests/test_batching.py
sabard: Explained offline—first select statement is to determine count.

graphene_sqlalchemy/batching.py
flipbit03: September update: There is also preliminary work in Graphene to vendor the `DataLoader` from `aiodataloader` into Graphene, and also change it's `__init__` behavior to not ask for an even loop so early on init. Just warning because that might touch the same points we are touching here.

sabard (approved):

erikwrede (approved):

erikwrede:
@sabard are we ready to merge?

PR #356, opened 2022-07-27
Draft: Add filter tests for discussion

sabard:
The purpose of this Draft PR is to facilitate discussion around implementing filters and to provide a concrete set of test cases that filtering must support in order to be incorporated into graphene-sqlalchemy.

So far, I have added test cases for the following:
- simple filter on field
- custom filter
- relationship filters (requires improvement)
- logic operators

I still need to implement test cases for the following:
- hybrid property

Please feel free to discuss general topics related to implementing filters here as well as specific improvements needed to testing and filter syntax.

### DIFF
--- a/graphene_sqlalchemy/filters.py
+++ b/graphene_sqlalchemy/filters.py
@@ -1,0 +1,2 @@
+class FloatFilter:
+    pass
--- a/graphene_sqlalchemy/tests/models.py
+++ b/graphene_sqlalchemy/tests/models.py
@@ -80,24 +80,9 @@ class Reporter(Base):
 articles_tags_table = Table(
     "articles_tags",
     Base.metadata,
-    Column("article_id", ForeignKey("article.id")),
-    Column("imgae_id", ForeignKey("image.id")),
+    Column("article_id", ForeignKey("articles.id")),
+    Column("tag_id", ForeignKey("tags.id")),
 ) 
-
-
-class Article(Base):
-    __tablename__ = "articles"
-    id = Column(Integer(), primary_key=True)
-    headline = Column(String(100))
-    pub_date = Column(Date())
-    reporter_id = Column(Integer(), ForeignKey("reporters.id"))
-
-    # one-to-one relationship with image
-    image_id = Column(Integer(), ForeignKey('image.id'), unique=True)
-    image = relationship("Image", backref=backref("articles", uselist=False))
-
-    # many-to-many relationship with tags
-    tags = relationship("Tag", secondary=articles_tags_table, backref="articles")
 
 
 class Image(Base):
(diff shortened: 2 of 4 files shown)


sabard:
@erikwrede not sure if I'm understanding your point about MultipleOfType correctly. Currently, the syntax would support this kind of query:
```
query {
  x (filters: {
    y: { name: {in: ["A", "B"]} }
  }) {
    ...
  }
}
```

SQLAlchemy: `sa.query(x).join(x.y).filter(y.name.in_("A", "B")) `

Is this kind of query you mentioned in slack? Where does MultipleOfType come in to support this?
```
query {
  x (filters: {
    and: [
      y: { name: {in: ["A", "B"]} },
      z: { id: 123 },
    ]
  }) {
    ...
  }
}
```

SQLAlchemy: `sa.query(x).join(x.y).filter(y.name.in_("A", "B")).join(x.z).filter(z.id=123) `

Let me know if I have anything wrong here.

erikwrede:
Hey, sorry for not following up with an example. Think of my comment like a friendsWith filter on User that has a list of friends. Currently you could only filter for "A is friends with one of B or C." What's missing is "A is friends with ALL of B and C" or   "A's friends are exactly B and C"

sabard:
Thanks for the detailed examples here. This makes a lot more sense to me and I'll update the tests to reflect the desired behavior of containsAllOf and containsExactly. Happy to aim for this support in the first release. Are there any other default RelationshipFilter behaviors we want to support? Maybe a better question is how can we make it easy for users to augment custom filters on top of this in the case they'd like to implement something like containsOneOf or containsNoneOf?

palisadoes:
Sqlalchemy has .startswith and .endswith that we should consider as
filtering options.

I discovered an issue with VARBINARY filtering with .startswith and
.endswith that should have been fixed. You may want to take a look just in
case as it may have a material impact:
https://github.com/sqlalchemy/sqlalchemy/issues/8253#issuecomment-1183453178

>

sabard:
Just added a couple subtests for a `RelationshipFilter` for 1:n and n:m relationshpis with `contains`, `containsAllOf`, and `containsExactly`. Let me know what you think. One question: Is it correct to provide roughly the same filter type for 1:n and n:m relationships? I assume something in the implementation will need to detect the exact relationship type and make sure the correct SQLAlchemy calls are made.

As for `startsWith` and `endsWith`, maybe we add examples of doing this as a `RelationshipFilter` extension for now? Would it be correct to assume that anything from [this RelationshipProperty list](https://docs.sqlalchemy.org/en/14/orm/internals.html?highlight=ilike#sqlalchemy.orm.RelationshipProperty) could potentially be fair game for providing as a default option on `RelationshipFilter`?

I'm going to go ahead and start implementing basic filtering on top of this PR rebased off #355, but will keep updating this as needed.

sabard:
Closing in favor of #357

PR #357, opened 2022-08-11
Add Filters

sabard:
Supersedes #356

### DIFF
--- a/graphene_sqlalchemy/filters.py
+++ b/graphene_sqlalchemy/filters.py
@@ -1,0 +1,2 @@
+class FloatFilter:
+    pass
--- a/graphene_sqlalchemy/tests/models.py
+++ b/graphene_sqlalchemy/tests/models.py
@@ -80,24 +80,9 @@ class Reporter(Base):
 articles_tags_table = Table(
     "articles_tags",
     Base.metadata,
-    Column("article_id", ForeignKey("article.id")),
-    Column("imgae_id", ForeignKey("image.id")),
+    Column("article_id", ForeignKey("articles.id")),
+    Column("tag_id", ForeignKey("tags.id")),
 ) 
-
-
-class Article(Base):
-    __tablename__ = "articles"
-    id = Column(Integer(), primary_key=True)
-    headline = Column(String(100))
-    pub_date = Column(Date())
-    reporter_id = Column(Integer(), ForeignKey("reporters.id"))
-
-    # one-to-one relationship with image
-    image_id = Column(Integer(), ForeignKey('image.id'), unique=True)
-    image = relationship("Image", backref=backref("articles", uselist=False))
-
-    # many-to-many relationship with tags
-    tags = relationship("Tag", secondary=articles_tags_table, backref="articles")
 
 
 class Image(Base):
@@ -111,6 +96,21 @@ class Tag(Base):
     __tablename__ = "tags"
     id = Column(Integer(), primary_key=True)
     name = Column(String(30))
+
+
+class Article(Base):
+    __tablename__ = "articles"
+    id = Column(Integer(), primary_key=True)
+    headline = Column(String(100))
+    pub_date = Column(Date())
+    reporter_id = Column(Integer(), ForeignKey("reporters.id"))
+
+    # one-to-one relationship with image
+    image_id = Column(Integer(), ForeignKey('images.id'), unique=True)
+    image = relationship("Image", backref=backref("articles", uselist=False))
+
+    # many-to-many relationship with tags
+    tags = relationship("Tag", secondary=articles_tags_table, backref="articles")
 
 
 class ReflectedEditor(type):
--- a/graphene_sqlalchemy/tests/models.py
+++ b/graphene_sqlalchemy/tests/models.py
@@ -6,7 +6,7 @@ from sqlalchemy import (Column, Date, Enum, ForeignKey, Integer, String, Table,
                         func, select)
 from sqlalchemy.ext.declarative import declarative_base
 from sqlalchemy.ext.hybrid import hybrid_property
-from sqlalchemy.orm import column_property, composite, mapper, relationship
+from sqlalchemy.orm import backref, column_property, composite, mapper, relationship
 
 PetKind = Enum("cat", "dog", name="pet_kind")
 
@@ -39,6 +39,7 @@ class Pet(Base):
     pet_kind = Column(PetKind, nullable=False)
     hair_kind = Column(Enum(HairKind, name="hair_kind"), nullable=False)
     reporter_id = Column(Integer(), ForeignKey("reporters.id"))
+    legs = Column(Integer(), default=4)
 
 
 class CompositeFullName(object):
@@ -76,12 +77,40 @@ class Reporter(Base):
     composite_prop = composite(CompositeFullName, first_name, last_name, doc="Composite")
 
 
+articles_tags_table = Table(
+    "articles_tags",
+    Base.metadata,
+    Column("article_id", ForeignKey("article.id")),
+    Column("imgae_id", ForeignKey("image.id")),
+) 
+
+
 class Article(Base):
     __tablename__ = "articles"
     id = Column(Integer(), primary_key=True)
     headline = Column(String(100))
     pub_date = Column(Date())
     reporter_id = Column(Integer(), ForeignKey("reporters.id"))
+
+    # one-to-one relationship with image
+    image_id = Column(Integer(), ForeignKey('image.id'), unique=True)
+    image = relationship("Image", backref=backref("articles", uselist=False))
+
+    # many-to-many relationship with tags
+    tags = relationship("Tag", secondary=articles_tags_table, backref="articles")
+
+
+class Image(Base):
+    __tablename__ = "images"
+    id = Column(Integer(), primary_key=True)
+    external_id = Column(Integer())
+    description = Column(String(30))
+
+
+class Tag(Base):
+    __tablename__ = "tags"
+    id = Column(Integer(), primary_key=True)
+    name = Column(String(30))
 
 
 class ReflectedEditor(type):
--- a/graphene_sqlalchemy/tests/test_filters.py
+++ b/graphene_sqlalchemy/tests/test_filters.py
@@ -1,4 +1,5 @@
 import graphene
+import pytest
 
 from ..fields import SQLAlchemyConnectionField
 from ..filters import FloatFilter
@@ -77,6 +78,7 @@ def create_schema(session):
 
 
 # Test a simple example of filtering
+@pytest.mark.xfail
 def test_filter_simple(session):
     add_test_data(session)
     Query = create_schema(session)
@@ -99,6 +101,7 @@ def test_filter_simple(session):
 
 
 # Test a custom filter type
+@pytest.mark.xfail
 def test_filter_custom_type(session):
     add_test_data(session)
     Query = create_schema(session)
@@ -132,6 +135,7 @@ def test_filter_custom_type(session):
     assert result == expected
 
 # Test a 1:1 relationship
+@pytest.mark.xfail
 def test_filter_relationship_one_to_one(session):
     article = Article(headline='Hi!')
     image = Image(external_id=1, description="A beautiful image.")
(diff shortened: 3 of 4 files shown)


graphene_sqlalchemy/filters.py
```diff
@@ -0,0 +1,180 @@
...
+
+
+class NumberFilter(FieldFilter):
+    """Intermediate Filter class since all Numbers are in an order relationship (support <, > etc)"""
+
+    class Meta:
+        abstract = True
+
+    @classmethod
+    def gt_filter(cls, val: AbstractType) -> bool:
```
sabard, line 165: Should this live on `NumberFilter`? Technically strings are ordered as well. Maybe we can call this `OrderedFilter` and numbers, strings, etc. can inherit?

erikwrede: Great point, added!

graphene_sqlalchemy/filters.py
```diff
@@ -44,10 +44,29 @@ def execute_filters(cls: Type[FieldFilter], query, filter_dict: Dict) -> Tuple[Q
...
-            clauses.extend(_clauses)
+            # Relationships are Dynamic, we need to resolve them fist
+            # Maybe we can cache these dynamics to improve efficiency
+            # Check with a profiler is required to determine necessity
+            input_field = cls._meta.fields[field]
+            if isinstance(input_field, graphene.Dynamic):
+                field_filter_type = input_field.get_type().type
+            else:
+                field_filter_type = cls._meta.fields[field].type
+            # TODO we need to save the relationship props in the meta fields array
```
erikwrede, line 55: We need to make sure to explicitly alias every joined model 
https://docs.sqlalchemy.org/en/14/changelog/migration_20.html#orm-query-join-aliased-true-from-joinpoint-removed

graphene_sqlalchemy/filters.py
```diff
@@ -0,0 +1,269 @@
...
+        # Pass modified meta to the super class
+        super(FieldFilter, cls).__init_subclass_with_meta__(_meta=_meta, **options)
+
+    # Abstract methods can be marked using AbstractType. See comment on the init method
+    @classmethod
+    def eq_filter(cls, query, field, val: AbstractType) -> Union[Tuple[Query, Any], Any]:
+        return field == val
+
+    @classmethod
+    def n_eq_filter(cls, query, field, val: AbstractType) -> Union[Tuple[Query, Any], Any]:
```
sabard, line 200: thoughts on using `ne_filter` to stay consistent with graphene-sqlalchemy-filter and sqlalchemy methods?

graphene_sqlalchemy/filters.py
```diff
@@ -0,0 +1,293 @@
...
+                # add function and attribute name to the list
+                filter_functions.append((re.sub("_filter$", "", func), get_type_hints(func_attr)))
+
+        relationship_filters = {}
+
+        # Generate Graphene Fields from the filter functions based on type hints
+        for field_name, _annotations in filter_functions:
+            assert "val" in _annotations, "Each filter method must have a value field with valid type annotations"
+            # If type is generic, replace with actual type of filter class
+            if is_list(_annotations["val"]):
```
erikwrede, line 114: reminder: cross-check with python list type hints of converter.py

palisadoes:
I know this is out of scope for this PR, but it would be good to add a short doc on the use of  `BatchSQLAlchemyConnectionField` with a brief comment. It's only documented in the test files, but should also be included in the documentation.

PaulSchweizer:
I tested the filters on our app and database and here are my first findings/questions:

- "contains_exactly" immediately runs in out of memory: `(MySQLdb._exceptions.OperationalError) (1038, 'Out of sort memory, consider increasing server sort buffer size')\n[SQL: SELECT count(*) ...` This query would only return 28 entities
- How would I the DateFilter for my datetime/date columns? Seems like the DateFilter is available but not registered

forana:
This PR has been our only blocker for upgrading for quite some time - how can we help move this forward?

sabard:
@forana I'm moving the `contains_exactly` implementation to a different PR so that the changes here can be merged in ASAP.

The main issue I'm still facing now that tests run both sync and async is differing async behavior when filtering over a relationship. As you can [see in this failing test](https://github.com/graphql-python/graphene-sqlalchemy/actions/runs/4659667382/jobs/8246805808?pr=357#step:5:310), the async tests is returning duplicate results.

@erikwrede or @PaulSchweizer do you all have any ideas about why this could be happening> I haven't had a chance to play around with async yet.

@PaulSchweizer will take a look at your`DateFilter` and `and/or` comments before the call today.

graphene_sqlalchemy/filters.py
```diff
@@ -0,0 +1,463 @@
...
+        # Generate Graphene Fields from the filter functions based on type hints
+        for field_name, _annotations in logic_functions:
+            assert (
+                "val" in _annotations
+            ), "Each filter method must have a value field with valid type annotations"
+            # If type is generic, replace with actual type of filter class
+
+            replace_type_vars = {BaseTypeFilterSelf: cls}
+            field_type = convert_sqlalchemy_type(
+                _annotations.get("val", str), replace_type_vars=replace_type_vars
```
PaulSchweizer, line 62: I investigated the issue I am experiencing where all the and/or filters turn out to be of type "[String]".
So far I was unable to reproduce this outside of our production setup but I have narrowed it down to this line, where the annotation["val"] turns out to be typing.List[String] for and/or. I am still trying to create a reproducible test, but maybe this info already sparks a thought that helps @sabard or @erikwrede

PaulSchweizer: Got it! As soon as you register custom types someting in the registry gets messed up and will pick the wrong type for and/or.

This will result in and/or becoming [String]:

```python
import sqlalchemy.types
from graphene_sqlalchemy.converter import convert_sqlalchemy_type
from graphene import String

@convert_sqlalchemy_type.register(
    sqlalchemy.types.PickleType
)
def convert_column_to_string(*args, **kwargs):
    return String
```

Too late now but hopefully this helps a bit more. Will try to setup a test sometime soon.

graphene_sqlalchemy/filters.py
```diff
@@ -0,0 +1,463 @@
...
+    @classmethod
+    def ilike_filter(cls, query, field, val: ScalarFilterInputType) -> bool:
+        return field.ilike(val)
+
+    @classmethod
+    def notlike_filter(cls, query, field, val: ScalarFilterInputType) -> bool:
+        return field.notlike(val)
+
+
+class BooleanFilter(FieldFilter):
```
palisadoes, line 309: This is preventing filtering on Boolean values. The filter expects a string such as "true", which doesn't exist in the DB. Using `true` causes a `"String cannot represent a non string value: true"` error

graphene_sqlalchemy/filters.py
```diff
@@ -0,0 +1,463 @@
...
+    @classmethod
+    def lt_filter(cls, query, field, val: ScalarFilterInputType) -> bool:
+        return field < val
+
+    @classmethod
+    def lte_filter(cls, query, field, val: ScalarFilterInputType) -> bool:
+        return field <= val
+
+
+class NumberFilter(OrderedFilter):
```
palisadoes, line 335: I think we need a `BooleanFilter` as well. Filtering on `graphene.Boolean` objects isn't possible now.

graphene_sqlalchemy/filters.py
sabard: Hm, I'm not able to reproduce this and boolean filters are working for me, but I'll throw it in a test case for completeness. What version of the project are you using?

forana:
I'm realizing that `graphene-sqlalchemy-filters` handled pagination, while this does not - any guidance for a recommended strategy for implementing pagination using these filters?

forana:
> > I'm realizing that `graphene-sqlalchemy-filters` handled pagination, while this does not - any guidance for a recommended strategy for implementing pagination using these filters?
> 
> Why would pagination not work if we're just attaching clauses to the SQLAlchemy query?

Not so much "this breaks pagination" as "this doesn't provide it, and I've realized I'll need to handle it" - I had been (incorrectly) thinking of this PR as a complete replacement for `graphene-sqlalchemy-filters` (and to be clear, I don't think it needs to be - just looking for recommendations/pointers before I roll my own pagination).

PaulSchweizer:
Pagination via offset and limit has to be implemented manually at this point, it could look something like this (going off of my memory):

```python
class PaginationConnectionField(graphene_sqlalchemy.SQLAlchemyConnectionField):
    def __init__(self,  *args, **kwargs):
        kwargs["page"] = graphene.Int()
        kwargs["itemsPerPage"] = graphene.Int()
        super().__init__(*args, **kwargs)

    @classmethod
    def get_query(cls, *args, **kwargs):
        query = SQLAlchemyConnectionField.get_query(*args, **kwargs)

        if (
            kwargs.get("page", None) is not None
            and kwargs.get("itemsPerPage", None) is not None
        ):
            query = query.offset(kwargs["page"] * kwargs["itemsPerPage"])
            query = query.limit(kwargs["itemsPerPage"])

        return query
```

We already briefly discussed implementing this (or a version of this) in the library, but there is no concrete plan yet.

PaulSchweizer:
We looked into the filters a bit more in depth now and came across this issue/question regarding filtering across relationships:

The new implementation uses INNERJOINs for the filters, `graphene_sqlalchemy_filters` on the other hand uses EXISTS statements to filter across relationships.

The graphql syntax of the two filter systems is basically the same. 

This lead to some initial confusion on our end since we were getting different results when translating our existing queries one to one into the new filter syntax.

I prepared an example to showcase the "issue", unfortunately I had no time yet to put it up somewhere but I can showcase it in a call. 

In the example we have Teams, which have a number of Tasks wich in turn have a number of Todos. Users are then assigned to the Todos (Teams -> Tasks -> Todos -> User).

This is the filter, translated 1:1 from the `graphene_sqlalchemy_filters` syntax:

```
query {
  teams(
    filter: {
      tasks: {
        contains: {
          and: [
            { name: { eq: "Accounting" } }
            { todos: { contains: { assignee_id: { eq: 2 } } } }
          ]
        }
      }
    }
  )
```

I would expect to get: 

`All teams that have an 'Accounting' task that contains a Todo where a user with id 2 is assigned.`

Instead we get: 

`All teams that EITHER have an 'Accounting' task OR a Task with a Todo where a user with id 2 is assigned.`

The "correct" query to get the initially expected result with the "new filters" would look like this:

```
query {
  teams(
    filter: {
      and: [
        {
          tasks: {
            contains: { name: { eq: "Accounting" } }
          }
        },
         {
          tasks: {
            contains: { todos: { contains: { assignee_id: { eq: 2 } } } }
          }
        }
      ]
    }
  )
```

BUT this one also does not work because each occurence of "tasks" gets a unique alias, so we again end up with the wrong result. The resulting SQL statement:

```
SELECT DISTINCT 
teams.id AS teams_id, 
FROM teams 
INNER JOIN tasks AS tasks_1 ON teams.id = tasks_1.shot_id 
INNER JOIN tasks AS tasks_2 ON teams.id = tasks_2.shot_id INNER JOIN todos AS todos_1 ON tasks_2.id = todos_1.task_id 
```

Should look like this:

```
SELECT DISTINCT 
teams.id AS teams_id, 
FROM teams 
INNER JOIN tasks AS tasks_1 ON teams.id = tasks_1.shot_id 
INNER JOIN todos AS todos_1 ON tasks_1.id = todos_1.task_id 
```

The solution here would probably be to use the same alias on each "level" of relationship.

This however highlights the question: Which behavior do we want to support? INNERJOINS or EXISTS? We talked about it in my team and came to the conclusion that the grapqhl filter syntax would rather suggest the EXISTS behavior over the INNERJOIN behaviour. Let me know what you think, probably best to jump on a call and discuss it in person (I would be free this coming Monday, May 15th)

palisadoes:
@sabard @erikwrede When do you think you'll be able to merge the filters into the main branch? Is it just the Enum work that is outstanding? I really need to migrate off the old graphene-sqlalchemy-filter.

erikwrede:
@palisadoes just fixed the enum filter case. AFAIK that covers our base scope, so what's missing now is a refactor and cleanup of the code, as it still contains many special cases from previous prototypes that can be generalized to improve code quality and readability.

palisadoes:
@erikwrede Sorry about missing the meeting. Are there any more outstanding PRs for the filtering?

PR #361, opened 2022-09-09
Add Black to pre-commit

erikwrede:
Adds black to pre-commit to ensure a consistent code style over all future contributions


sabard (changes requested):
Possible to change:
```Python
default_language_version:
  python: python3.10
```
to 
```Python
default_language_version:
  python: python3
```
in .pre-commit-config.yaml? I'm using Python3.8 now and I can't lint without that. I'd make the change, but don't have access to your repo.

The other comments here are just my preference; let me know what you think. Otherwise looks great.

sabard (approved):

PR #362, opened 2022-09-12
Fix: RelationshipLoader should utilize DataLoader from the `get_data_loader_impl()` function, and not directly from aiodataloader library

flipbit03:


graphene_sqlalchemy/batching.py
```diff
@@ -2,15 +2,28 @@
...
 
-import aiodataloader
 import sqlalchemy
 from sqlalchemy.orm import Session, strategies
 from sqlalchemy.orm.query import QueryContext
 
 from .utils import is_graphene_version_less_than, is_sqlalchemy_version_less_than
 
 
-class RelationshipLoader(aiodataloader.DataLoader):
```
flipbit03, line 13: RelationshipLoader is pointing to `aiodataloader.DataLoader`, when it should be using the locally-sourced DataLoader that comes from the `get_data_loader_impl()` function.

flipbit03 (commented):
@erikwrede Thanks in advance!

erikwrede (approved):

PR #364, opened 2022-10-19
Make column not required if server_default exists

0x1723:
This PR does the following:

- A function called `get_column_server_default` which gets the given column's `server_default` value from the schema definition.
- Only sets the graphql field as required if it is `nullable` and doesn't have a `server_default value`.

This is because when there is a `server_default` set, we might not wanna have it be required, and have the `server_default` value be used when creating a new entry.

### DIFF
--- a/graphene_sqlalchemy/converter.py
+++ b/graphene_sqlalchemy/converter.py
@@ -54,6 +54,10 @@ def is_column_nullable(column):
     return bool(getattr(column, "nullable", True))
 
 
+def get_column_server_default(column):
+    return getattr(column, "server_default", None)
+
+
 def convert_sqlalchemy_relationship(
     relationship_prop,
     obj_type,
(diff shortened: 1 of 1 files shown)


0x1723:
@jnak would this addition be good? If so, I'm a first-time contributor so this would need a maintainer to approve running workflows.

erikwrede:
hey, sorry for the long wait!
I'm not sure if I understand the purpose of this PR correctly. Just because a `column` has a `server_default`, it doesn't mean that updates to it can set it to `null`, causing a `null` return for a required field. The `server_default` is only relevant during the insert operation; it translates 

```python
Column('test', DateTime, server_default=text('NOW()'))
```
to
```sql
test DATETIME DEFAULT NOW()
```

In fact, setting the default to `null` is also supported. Conclusively, making the required status of a field depend on the `server_default` may yield unexpected behavior.

erikwrede:
Hey @0x1723,
I'm closing this for now due to the reasons stated above. If this comes up again, feel free to post an issue referencing this PR. Happy to elaborate further over there. ☺️

Cheers
Erik

PR #365, opened 2022-10-24
Support GQL interfaces for polymorphic SQLA models

polgfred:
Hi!

Would love any feedback on this.

My background assumption here is that, for most folks using `graphene_sqlalchemy` bindings, GraphQL interfaces are going to take the form of polymorphic base classes. So it would be great if you could extend `SQLAlchemyInterface` the same way you extend `SQLAlchemyObjectType` for concrete types -- and wire up the attributes using the same reflection mechanism.

So, in order to accomplish this, I:
- moved the logic in `SQLAlchemyObjectType` up into a `SQLAlchemyBase` type
- made `SQLAlchemyObjectType` and (a new) `SQLAlchemyInterface` extend `SQLAlchemyBase`
- updated `Registry` to accept objects of type `SQLAlchemyBase`

I've tested this in a work project that depends heavily on a few polymorphic JTI base types, and this mechanism works perfectly. However, I'm sure I haven't thought through potential limitations and incorrect assumptions, which is why I'm submitting this for review and feedback.

Thanks!


flipbit03 (approved):
LGTM!

erikwrede:
Thanks for the PR! Will check it out this week 🙂

erikwrede (changes requested):
Sorry for the delay, this looks great!

Can you please add a usage example, unit tests and some docstrings to give some guidance on when and how to use the new `Base` and `Interface` types?

fred-asimov:
> Can you please add a usage example, unit tests and some docstrings to give some guidance on when and how to use the new `Base` and `Interface` types?

@erikwrede Thanks! And sorry for the delay, been busy over here. I've added a [doc](docs/inheritance.rst), unit tests, and a docstring for `ObjectType` and `Interface`. Let me know if this is what you're looking for, and please feel free to tweak if desired.

polgfred:
> Can you please add a usage example, unit tests and some docstrings to give some guidance on when and how to use the new `Base` and `Interface` types?

@erikwrede Thanks! And sorry for the delay, been busy over here. I've added a doc, unit tests, and a docstring for `ObjectType` and `Interface`. Let me know if this is what you're looking for, and please feel free to tweak if desired.

erikwrede:
@polgfred thanks for the update! Appreciate the weekend-effort 🙂 

There's two things we should discuss before merging this:

**1. Should we explicitly exclude the field linked in `polymorphic_on` (in this case `type`) from the GQL schema?** 
The `type` field is verbose as GraphQL already provides the `__typename` field including the GQL type name. API-Design wise the type field does not bring any value as it is only relevant for server internals.
We could enable overriding the behavior using the `ORMField`: if an `ORMField` was created on the base/explicit type for the field containing the polymorphic identity, the field would not be excluded. 
I think this leads to better schema design while still giving users the option to customize.

**2. Removing `polymorphic_identity: "person"` from the examples **
`PersonType` is an `Interface` in this case, so pure instances of `PersonType` are not allowed. However, the underlying DB Model allows pure instances of our base class `Person`  because it provides a `polymorphic_identity`. This can lead to the following error if a pure `Person` instance is returned from the resolvers:

```python
 "message": "Abstract type 'PersonType' must resolve to an Object type at runtime for field 'Query.people'. Either the 'PersonType' type should provide a 'resolve_type' function or each possible type should provide an 'is_type_of' function."
```
If the `polymorphic_identity` is not set for `Person`, SQLAlchemy will not allow such inserts.

We need to clarify that the `Interface` type is only suitable for models that have an _abstract_ (maybe there's better wording) base that is guaranteed to not appear in the records.
Same goes for multi-table inhertance with a distinct table for the base (which is an uncommon case but it might still be used in existing designs).

polgfred:
@erikwrede Yeah, I agree with both those observations. An interface isn't _exactly_ a polymorphic type for the reason you mentioned, but in my experience it's much more common for a polymorphic base class to be abstract, so we should call that out in the docs and examples. And I agree that it makes sense to hide the `polymorphic_on` field in the schema. I'll look into that tonight :)

polgfred:
@erikwrede Added checks for the stuff you called out.

erikwrede (approved):
Looks good now! Performed slight modifications to docs & the assertion error in SQLAlchemyInterface & added 3 unit tests. LMKWYT 🙂

polgfred:
Awesome, looks great to me!

graphene_sqlalchemy/types.py
```diff
@@ -345,3 +362,105 @@ def enum_for_field(cls, field_name):
...
+        )
+
+        # make sure that the model doesn't have a polymorphic_identity defined
+        if hasattr(_meta.model, "__mapper__"):
+            polymorphic_identity = _meta.model.__mapper__.polymorphic_identity
+            assert (
+                polymorphic_identity is None
+            ), '{}: An interface cannot map to a concrete type (polymorphic_identity is "{}")'.format(
+                cls.__name__, polymorphic_identity
+            )
```
erikwrede, line 466: @polgfred I just realized that this might have been a mistake to add. Sometimes we have a base type that is extended by other types but has a polymorphic identity. In this case, there would be an object type implementing the interface with no additional fields. We might need to revert hat bit, what do you think?

PR #367, opened 2022-11-28
Stricter non-null fields for relationships

polgfred:
Following on the heels of my first successful PR 😉, I wanted to float another suggestion for your consideration. It seems like it would be closer to ORM semantics if:

- For to-one relationships, if the ID field (e.g. `user_id`) is non-null, the relationship field should be also
- For to-many relationships, both the list element type and the list field itself should be non-null, so that there is always a list for a many relationship (even if it is empty), and it never contains None

For example:

```python
    class User(Base):
        comments = relationship("Comment")

    class Comment(Base):
        user_id = Column(Integer(), nullable=False)
        user = relationship("User")
```

The Schema should be:

```graphql
    type User {
        comments: [Comment!]!
    }

    type Comment {
        user_id: Int!
        user: User!
    }
```

This should probably be the default (version 3 maybe?), but in the interests of backward compatibility, we can make it an opt-in feature. This would help my UI team immensely by cleaning up the schema, and not having to deal with unnecessary `null`s all over the place.

Obviously more tests and docs would be forthcoming, but I wanted to float the idea first.

cc: @erikwrede @flipbit03


flipbit03:
I really like this idea and I think it makes sense, given that the SQLAlchemy relationships (be it a vector or a scalar) always return the `T` inside them, there's not really any `None`'s in those.

erikwrede:
This really makes sense.
But we need another option to check for nullability of :1 relationships first. The method you used isn't guaranteed to work (just like you mentioned in your comment).

Additionally, we should also change Relay's behavior (connectionFields and nodes) to keep everything nice and consistent.

Another concern I have is future support of the @defer directive which is going to be released in graphql-js soon, meaning that graphql-python will follow. At some point, deferring was only supported on nullable fields. Since deferring makes a lot of sense on relationships to improve response times, we should see if the new spec might impact the design in graphene-sqlalchemy.

Will have a more detailed look at that this week!

erikwrede:
I haven't found an elegant approach to the `to-one` relationship nullability check so far. We could use the column containing the foreign key and check `column.nullable`. However, I  couldn't find a way to get that column name from the relationship property, other than manually parsing the `A.primary_key == B.foreign_key` expression in the relationship prop. Any ideas?

polgfred:
> We should discuss if it makes sense to include the to-one relationships in this PR if there is no clearer approach for both sides of the relationship.

@erikwrede Yeah, the to-many relationship scenario is much more straightforward. I'd be ok punting on that. I haven't dug into it more than I did in the original PR, but what I came up with was unfortunately the most elegant approach I could come up with. :)

erikwrede:
Okay, let's push back the to-one case and focus on to-many relationships here! 

We're still missing
- [ ] Test Coverage for the newly added lines
- [ ] Additional tests to check the generated types in both cases (NonNull, Nullable)

When that is done, I'll merge it for 3.0 🙂

polgfred:
> When that is done, I'll merge it for 3.0

So in that case, should I remove the `use_non_nullable_relationships` flag, and just make this behavior the default?

erikwrede:
I believe we should make the `NonNull` behavior the default but still give users the choice in case they want to be fully relay-spec compliant. The flag is a great compromise for that.

erikwrede (approved):
LGTM! Thanks for getting this started 🙂

polgfred:
> LGTM! Thanks for getting this started 🙂

Thanks! I wasn't sure if there were some other more comprehensive unit tests you wanted, but I added ones to cover the changes.

erikwrede:
The logic behind this change is quite simple and there shouldn't be any complex collisions, so making change the new behavior works and the legacy behavior should suffice in this case. 😊

PR #368, opened 2022-12-04
Draft: featSQLAlchemy 2.0

erikwrede:


erikwrede:
Some `Exception ignored` errors have appeared on 2.0:
```sh
Exception ignored in: <function _ConnectionRecord.checkout.<locals>.<lambda> at 0x11159d7a0>
  File "/Users/erik/dev/graphene-sqlalchemy/venv/lib/python3.7/site-packages/sqlalchemy/pool/base.py", line 734, in <lambda>
    if _finalize_fairy is not None
  File "/Users/erik/dev/graphene-sqlalchemy/venv/lib/python3.7/site-packages/sqlalchemy/pool/base.py", line 1030, in _finalize_fairy
    fairy.dbapi_connection = fairy._connection_record = None  # type: ignore
AttributeError: 'NoneType' object has no attribute 'dbapi_connection'
```
These seem to be related to  the pytest fixtures, not the actual test cases. I haven't been able to identify a cause so far.

graphene_sqlalchemy/tests/test_converter.py
```diff
@@ -336,6 +337,22 @@ class TestEnum(enum.IntEnum):
     assert graphene_type._meta.enum.__members__["two"].value == 2
 
 
+@pytest.mark.skipif(
+    not SQL_VERSION_HIGHER_EQUAL_THAN_1_4,
+    reason="SQLAlchemy <1.4 does not support this",
+)
+def test_should_columproperty_convert_sqa_20():
```
jendrikjoe, line 344: NIT: You could hand the select function and parametrize the test depending on the SQL version 🙂

graphene_sqlalchemy/tests/models_batching.py
```diff
@@ -60,9 +62,14 @@ class Reporter(Base):
...
-    column_prop = column_property(
-        select([func.cast(func.count(id), Integer)]), doc="Column property"
-    )
+    if SQL_VERSION_HIGHER_EQUAL_THAN_1_4:
+        column_prop = column_property(
+            select(func.cast(func.count(id), Integer)), doc="Column property"
+        )
+    else:
+        column_prop = column_property(
+            select([func.cast(func.count(id), Integer)]), doc="Column property"
```
jendrikjoe, line 71: Actually as you are doing the `select(...)` vs. `select([...])` a lot, maybe a helper function:
```python
def wrap_select_queries(query):
    if SQL_VERSION_HIGHER_EQUAL_THAN_1_4:
        return select(query)
    else:
        return select([query])
```
might be useful 🙂 
Makes dropping the <1.4 support later a lot easier as you would just need to search for `wrap_select_queries` and replace it with select 🙂

jendrikjoe (changes requested):
Rest seems great to me 👍

graphene_sqlalchemy/tests/test_converter.py
erikwrede: Good Idea!

pappasam:
@erikwrede just checking in to see if there's any help you need here to get this to the finish line? I'm not sure if this is paused because of some other piece of dependent work, but I'm very excited to use `sqlalchemy` 2.0 in my applications and this (AFAIK) is my final blocker.

erikwrede:
@pappasam on this branch, all tests are passing on 2.0. Fixing the review comments is still on my list and will be done soon. You can already start testing by installing this branch via pip. Async sessions are also supported already. Would appreciate any feedback! :)

pappasam:
@erikwrede I can confirm that this branch works for me! No issues I can find on my end; I'm very much looking forward to this making it into a beta release

Issue #102, opened 2017-12-17
Automatic conversion from Int to ID is problematic


Cito:
Graphene-SQLAlchemy automatically converts columns of type SmallInteger or Integer to `ID!` fields if they are primary keys, but does not convert such columns to `ID` fields if they are foreign keys.

Take for example this schema:
```python
class Department(Base):
    __tablename__ = 'department'
    id = Column(Integer, primary_key=True)
    name = Column(String)

class User(Base):
    __tablename__ = 'users'
    id = Column(Integer, primary_key=True)
    name = Column(String)
    department_id = Column(Integer, ForeignKey('department.id'))
    department = relationship(Department)

class DepartmentType(SQLAlchemyObjectType):
    class Meta:
        model = Department

class UserType(SQLAlchemyObjectType):
    class Meta:
        model = User

class Query(ObjectType):

    departments = List(DepartmentType)
    users = List(UserType)

    def resolve_departments(self, info):
        return DepartmentType.get_query(info)

    def resolve_users(self, info):
        return UserType.get_query(info)
```
You can run the following query:
```graphql
query {
  users {
    id
    name
    departmentId
    department {
      id
    }
  }
}
```
As a result, you get something like:
```json
{
  "data": {
    "users": [
      {
        "id": "1",
        "firstName": "Fred",
        "departmentId": 1,
        "department": {
          "id": "1"
        }
      },
      {
        "id": "2",
        "firstName": "Barnie",
        "departmentId": 2,
        "department": {
          "id": "2"
        }
      }
    ]
  }
}
```
As you see, `departmend.id` is a string (because IDs are returned as strings), while `departmentId` is a number. This turned out to be a huge problem and source of error in practice. Working with this inconsistent, fault-prone interface has bitten me many times. When storing ids in objects on the frontend, or using ids as filters, I never know whether I should use numbers or strings. Currently I have conversions from number to string and vice versa everywhere in my frontend code, and if I don't do it correctly, things stop working in hard to debug ways because you often don't recognize such type mismatches. On the server side, do I take ids as filter parameters  as IDs or Ints? If I do the former, I must then convert them to integer when using them as filter arguments for SQLAlchemy. So, really, this is no fun to work with and doesn't work in practice, because you always have this mental burden of thinking about whether your ids should be represented as strings or numbers and whether you need to convert them when passing them around.

[...]

PR #371, opened 2022-12-25
refactor!: use the same conversion system for hybrids and columns

erikwrede:
This PR refactors the type conversion system to use the same converters for all column types (column + hybrid). Previously, we had `convert_sqlalchemy_column` using `singledispatch` and `convert_sqlalchemy_hybrid_property_type` using a custom singledispatch-like solution to match column and hybrid property type(-hints) to their graphene equivalents. This implied having to maintain two separate type conversion systems with very similar functionality. This PR proposes merging the systems based on our custom singledispatch-style approach. This has several implications:

### Breaking Change: convert_sqlalchemy_type now uses a matcher function
Our custom matcher needs to be able to check more than just the type of a column. Some type hints like unions or optionals need further processing. To convert old singledispatch-style converters to the new converters, the column type has to be wrapped in our custom `value_equals` helper function:

```
# Before 
@convert_sqlalchemy_type.register(JSON)

# After
@convert_sqlalchemy_type.register(value_equals(JSON))
```
### Breaking Change: convert_sqlalchemy_type support for subtypes is dropped, each column type must be explicitly registered
Since we now use matcher functions, it doesn't make sense to check if any inherited types match the singledispatch converters, as they might interfere with other matchers. Because of that, all types must now be explicitly registered:

```
@convert_sqlalchemy_type.register(value_equals(MYColumnType))
def convert_my_type(column_type, **kwargs):
   return MyGrapheneType
```

See `converter.py` for more detailed examples.

### Breaking Change: The first argument to converter functions is always a type
Previously, both type and instances of types were passed to the converter. (`String` vs `String(30)`). If you need information from the instance, use `column.type` instead.  *Remember to check if `column` is present and not a `hybrid_property`*

#### Example:
```python
@convert_sqlalchemy_type.register(column_type_eq(sqa_types.ARRAY))
@convert_sqlalchemy_type.register(column_type_eq(postgresql.ARRAY))
def convert_array_to_list(
    type_arg: Any,
    column: Optional[Union[MapperProperty, hybrid_property]] = None,
    registry: Registry = None,
    **kwargs,
):
    if column is None or isinstance(column, hybrid_property):
        raise Exception("SQL-Array conversion requires a column")
    item_type = column.type.item_type
    if not isinstance(item_type, type):
        item_type = type(item_type)
    inner_type = convert_sqlalchemy_type(
        item_type, column=column, registry=registry, **kwargs
    )
    return graphene.List(
        init_array_list_recursive(inner_type, (column.type.dimensions or 1) - 1)
    )
```
### Breaking Change: convert_sqlalchemy_type's column and registry arguments must now be keyword arguments
To make additional support for type conversions possible, we converted all arguments to the converters to keyword arguments. It is not guaranteed for these to be present, so you should always check for existance of `column` or `registry` if you plan on using them. 
Since other arguments might be provided as well, make sure that `**kwargs` is present on your signature and that it is passed to any calls of nested type conversions (e.g. `List[JSON]` calls `convert_list`, then `convert_list` starts another conversion for the inner type)

### Breaking Change: The hybrid property default column type is no longer a string. If no matching column type was found, an exception will be raised.
We introduced warnings for the string fallback in `3.0.0b2` as a soft transition to the new system. Since we merged both conversion systems, it is impractical to have a fallback for all types. If you want to keep the old `string` types for your hybrids, please use:

```python
class MyType(SQLAlchemyObjectType):
    class Meta:
        model = MyModelWithHybrid
    myHybrid = ORMField(type_=graphene.String)
```

### DIFF
--- a/graphene_sqlalchemy/converter.py
+++ b/graphene_sqlalchemy/converter.py
@@ -2,14 +2,13 @@ import datetime
 import sys
 import typing
 import uuid
-import warnings
 from decimal import Decimal
-from functools import singledispatch
-from typing import Any, cast
+from typing import Any, Optional, Union, cast
 
 from sqlalchemy import types as sqa_types
 from sqlalchemy.dialects import postgresql
-from sqlalchemy.orm import interfaces, strategies
+from sqlalchemy.ext.hybrid import hybrid_property
+from sqlalchemy.orm import MapperProperty, interfaces, strategies
 
 import graphene
 from graphene.types.json import JSONString
@@ -17,7 +16,7 @@ from graphene.types.json import JSONString
 from .batching import get_batch_resolver
 from .enums import enum_for_sa_enum
 from .fields import BatchSQLAlchemyConnectionField, default_connection_field_factory
-from .registry import get_global_registry
+from .registry import Registry, get_global_registry
 from .resolvers import get_attr_resolver, get_custom_resolver
 from .utils import (
     DummyImport,
@@ -25,6 +24,7 @@ from .utils import (
     safe_isinstance,
     singledispatchbymatchfunction,
     value_equals,
+    value_equals_strict,
 )
 
 try:
@@ -210,7 +210,9 @@ def convert_sqlalchemy_column(column_prop, registry, resolver, **field_kwargs):
 
     field_kwargs.setdefault(
         "type_",
-        convert_sqlalchemy_type(getattr(column, "type", None), column, registry),
+        convert_sqlalchemy_type(
+            getattr(column, "type", None), column=column, registry=registry
+        ),
     )
     field_kwargs.setdefault("required", not is_column_nullable(column))
     field_kwargs.setdefault("description", get_column_doc(column))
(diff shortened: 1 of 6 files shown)


graphene_sqlalchemy/utils.py
```diff
@@ -196,18 +196,23 @@ def __call__(self, *args, **kwargs):
...
 
-    def register(self, matcher_function: Callable[[Any], bool]):
-        def grab_function_from_outside(f):
-            self.registry[matcher_function] = f
-            return self
+    def register(self, matcher_function: Callable[[Any], bool], func=None):
+        if func is None:
+            return lambda f: self.register(matcher_function, f)
+        self.registry[matcher_function] = func
+        return func
```
erikwrede, line 203: Had to adapt this to functools' SingleDispatch to allow for directive chaining

graphene_sqlalchemy/utils.py
```diff
@@ -196,18 +196,23 @@ def __call__(self, *args, **kwargs):
...
 
-        return grab_function_from_outside
+
+def value_equals(value: Any) -> Callable[[Any], bool]:
+    """A simple function that makes the equality based matcher functions for
+    SingleDispatchByMatchFunction prettier"""
+    return lambda x: (x == value or type(x) is value)
```
erikwrede, line 209: Per default we now also check for isinstance. We should probably rename this to column_type_equals because it cannot be used for the builtins (see `str` + `ForwardRef`)

graphene_sqlalchemy/tests/models.py
```diff
@@ -254,9 +254,11 @@ def hybrid_prop_shopping_cart_item_list(self) -> List[ShoppingCartItem]:
...
 
     # Unsupported Type
-    @hybrid_property
-    def hybrid_prop_unsupported_type_tuple(self) -> Tuple[str, str]:
-        return "this will actually", "be a string"
+    # fixme move this somewhere else
+    #
+    # @hybrid_property
+    # def hybrid_prop_unsupported_type_tuple(self) -> Tuple[str, str]:
+    #     return "this will actually", "be a string"
```
erikwrede, line 261: Will add a separate test case for this before merging.

flipbit03 (dismissed):
LGTM! :shipit:

graphene_sqlalchemy/converter.py
```diff
@@ -207,97 +214,181 @@ def inner(fn):
...
 
 
-@convert_sqlalchemy_type.register(sqa_types.Float)
-@convert_sqlalchemy_type.register(sqa_types.Numeric)
-@convert_sqlalchemy_type.register(sqa_types.BigInteger)
-def convert_column_to_float(type, column, registry=None):
+@convert_sqlalchemy_type.register(column_type_eq(float))
+@convert_sqlalchemy_type.register(column_type_eq(sqa_types.Float))
+@convert_sqlalchemy_type.register(column_type_eq(sqa_types.Numeric))
+@convert_sqlalchemy_type.register(column_type_eq(sqa_types.BigInteger))
```
sabard, line 346: maybe a bit off topic here, but wondering about the choice of BigInteger->Float. assume this would break down after 64 bits. any plans for graphql-core to provide a [BigInt type like this](https://www.npmjs.com/package/graphql-bigint)?

sabard (commented):
nice—much cleaner! looking forward to seeing the rest of the changes.

graphene_sqlalchemy/tests/test_converter.py
```diff
@@ -78,6 +79,65 @@ def prop_method() -> int:
...
+    ):
+        get_hybrid_property_type(hybrid_prop)
+
+
+def test_hybrid_invalid_forward_reference():
+    class MyTypeNotInRegistry:
+        pass
+
+    @hybrid_property
+    def hybrid_prop(self) -> "MyTypeNotInRegistry":
```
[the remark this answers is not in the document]
erikwrede: Good point, added that!

graphene_sqlalchemy/converter.py
```diff
@@ -207,97 +214,181 @@ def inner(fn):
...
+def convert_sqlalchemy_type(
+    type_arg: Any,
+    column: Optional[Union[MapperProperty, hybrid_property]] = None,
+    registry: Registry = None,
+    **kwargs,
+):
+    registry_ = registry or get_global_registry()
+    # Enable returning Models by getting the corresponding model from the registry
+    # ToDo Move to a separate converter later, matching all sqlalchemy mapped
+    #  types and returning a dynamic
```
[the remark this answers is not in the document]
erikwrede: Yes, I moved it to a function now but will change it to a dynamic later so we can recognize and wrap it in another dynamic in the filter PR - similar to how we do it with relationship filters.

graphene_sqlalchemy/converter.py
```diff
@@ -207,97 +214,181 @@ def inner(fn):
...
+@convert_sqlalchemy_type.register(column_type_eq(sqa_types.SmallInteger))
+@convert_sqlalchemy_type.register(column_type_eq(sqa_types.Integer))
+@convert_sqlalchemy_type.register(column_type_eq(int))
+def convert_column_to_int_or_id(
+    type_arg: Any,
+    column: Optional[Union[MapperProperty, hybrid_property]] = None,
+    registry: Registry = None,
+    **kwargs,
+):
+    # fixme drop the primary key processing from here in another pr
```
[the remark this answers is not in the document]
erikwrede: Exactly what was mentioned in #102 and also the discussion we recently had about https://github.com/graphql-python/graphene/pull/1428
Out of scope for this refactor though, just added the fixme so it's marked in the code

graphene_sqlalchemy/converter.py
erikwrede: Good point, let's put this into an issue. Will have a look at it later.

graphene_sqlalchemy/converter.py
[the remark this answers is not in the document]
erikwrede: Ok I realized we can't use Dynamic here, because graphene can't unpack nested dynamics (e.g. `graphene.List(graphene.Dynamic)`)

So we'll need to check if a field type is a function (`inspect.isfunction(_type) or isinstance(_type, partial)`) additionally to the `instanceof dynamic`check in the filtering PR. This way we still maintain lazy support here.

graphene_sqlalchemy/converter.py
erikwrede: Started #373

erikwrede:
@sabard All fixed up, feel free to merge if nothing else comes up :)

sabard (approved):
lgtm!

guillaumep:
I've had regression in my code after upgrading graphene-sqlalchemy and it just by reading the changes one by one that I saw these breaking changes. I'd like to suggest such breaking changes to be indicated in the release notes otherwise they are hard to find.

erikwrede:
@guillaumep sorry about that, it seems like the release notes for this release were not published properly. We will fix that up soon. 
As a tip: in the change list, we will use semantic commits in the future, and in the notes of this release, this PR was already marked with a `!`, which suggests a breaking change. Of course I appreciate that this should not be the standard, and as you've seen from the other release notes, we usually publish more detailed information. For the 3.0 full release, there will be a full upgrade guide.

Apart from that, how is this change working for you? We haven't received too much feedback yet, so I'd love to hear your thoughts.

guillaumep:
> sorry about that, it seems like the release notes for this release were not published properly. We will fix that up soon.

Don't worry about it, and thanks for listening to my feedback.

> As a tip: in the change list, we will use semantic commits in the future, and in the notes of this release, this PR was already marked with a `!`, which suggests a breaking change.

I was not completely aware of this convention in open source projects. I'll pay more attention to it starting now.

> For the 3.0 full release, there will be a full upgrade guide.

Great!
 
> Apart from that, how is this change working for you? We haven't received too much feedback yet, so I'd love to hear your thoughts.

I upgraded from 3.0.0b3 to 3.0.0b4. Once I figured out how to change my code everything worked pretty well. We register a custom multi lingual dictionary SQLAlchemy type, which serializes data to a string in the database. We are providing the data as an object to the API clients. Here's the code diff for the upgrade:

```diff
from graphene_sqlalchemy.converter import convert_sqlalchemy_type
+from graphene_sqlalchemy.utils import column_type_eq

-@convert_sqlalchemy_type.register(ModelMlText)
+@convert_sqlalchemy_type.register(column_type_eq(ModelMlText))
def convert_mltext_to_json(type, column, registry=None):
    return MlText
```

The key here was to figure-out the use of `column_type_eq`. I read graphene-sqlalchemy's source code to see existing examples to figure this one out.

Thanks for the ongoing work on the project!

PR #372, opened 2023-01-01
Fix installation instruction: '--pre' instead of '>=3'

zahrevsky:
I didn't update version number as required by contrib guidelines, as this change affects only README.md, and doesn't change code behavior. Is it OK?

**About the change:**
Install instructions in the README.md fails with an error: „Could not find a version that satisfies the requirement graphene-sqlalchemy>=3“

This is because v3 is in beta. Therefore, installing with '--pre' fixes the problem.

Also, it's not a unique case — the same problem appears to be in `graphene-django`: https://github.com/graphql-python/graphene-django/issues/1288


erikwrede:
Thank you!

PR #376, opened 2023-01-13
Fix decimal type conversion

jendrikjoe:
With the merge of #371, the conversion of Decimal columns broke for us. 
I added a decimal column as we are creating them to the test model reproducing it and a possible fix, by registering the type to the converter.

### DIFF
--- a/graphene_sqlalchemy/converter.py
+++ b/graphene_sqlalchemy/converter.py
@@ -518,6 +518,7 @@ def convert_variant_to_impl_type(
     )
 
 
+@convert_sqlalchemy_type.register(column_type_eq(sqa_types.DECIMAL))
 @convert_sqlalchemy_type.register(column_type_eq(Decimal))
 def convert_sqlalchemy_hybrid_property_type_decimal(type_arg: Any, **kwargs):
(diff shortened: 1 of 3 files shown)


erikwrede:
Sorry for the late reply. 
I have received multiple messages about this, so I am considering making changes to the type conversion system introduced in #371 so that we keep supporting the conversion of subclasses of registered classes. This adds a lot of convenience since every driver has custom subtypes for bigint, integer and others.
Please check out #377 to see if that fixes your issues and give me some feedback ☺️ 
I will add the necessary tests tomorow.

jendrikjoe:
> Sorry for the late reply.
> I have received multiple messages about this, so I am considering making changes to the type conversion system introduced in #371 so that we keep supporting the conversion of subclasses of registered classes. This adds a lot of convenience since every driver has custom subtypes for bigint, integer and others.
> Please check out #377 to see if that fixes your issues and give me some feedback ☺️
> I will add the necessary tests tomorow.

No worries. Solving this in a generic way definitely seems reasonable 👍 Will close this one in favor of #377 👍

PR #378, opened 2023-01-25
add filters alias support

erikwrede:


graphene_sqlalchemy/filters.py
```diff
@@ -158,7 +187,9 @@ def execute_filters(
                 )
                 clauses.extend(_clauses)
             else:
-                model_field = getattr(model, field)
+                model_field = getattr(
+                    model, input_field.model_attr
+                )  # getattr(model, field)
```
sabard, line 192: Is it possible for `model_field` to be `None` here? Looking at [L#486 of types.py](https://github.com/graphql-python/graphene-sqlalchemy/blob/322f12a1523b4866bbc01bc21724281c8707d055/graphene_sqlalchemy/types.py#L486), couldn't a `graphene.InputField` be passed in to a `BaseTypeFilter`? Not sure if I understand the scope of this change though.

graphene_sqlalchemy/filters.py
```diff
@@ -35,6 +36,33 @@ def _get_functions_by_regex(
     return matching_functions
 
 
+class SQLAlchemyFilterInputField(graphene.InputField):
+    def __init__(
+        self,
+        type_,
+        model_attr,
+        name=None,
```
sabard, line 44: any consequences of having this be `None` for the superclass?

sabard (approved):

erikwrede:
Manually merged into add_filters

PR #379, opened 2023-01-27
feat!: configurable key to ID conversion

erikwrede:


graphene_sqlalchemy/converter.py
```diff
@@ -93,6 +93,14 @@ def set_non_null_many_relationships(non_null_flag):
     use_non_null_many_relationships = non_null_flag
 
 
+use_id_type_for_keys = True
```
jendrikjoe, line 96: NIT: As you have a setter for this, I would make it protected 😉

jendrikjoe (approved):
Rest seems good. Maybe adding a test to ensure that actually querying an object returns as well the expected type (e.g. int not the converted id) might be helpful just for completeness 👍

graphene_sqlalchemy/tests/test_converter.py
```diff
@@ -381,12 +383,28 @@ def test_should_integer_convert_int():
     assert get_field(types.Integer()).type == graphene.Int
 
 
-def test_should_primary_integer_convert_id():
+def test_should_key_integer_convert_id():
```
sabard, line 386: might be worth testing non-default behavior of setting `use_id_type_for_keys` to false too

PR #383, opened 2023-02-15
fix: warnings in docs build

sabard:
Small fixes to docs to eliminate compile warnings.
As of now, the custom sphinx theme does not properly format ``note`` blocks. Might be something to take a look at in the future.


sabard:
@erikwrede sorry, rebased. mind approving again?

Issue #35, opened 2017-02-16
N + 1 round trip problem


itaied246:
Does this library handle nested models (joins) in a single query from the server to the DB?
 For example

user {
  id
  posts {
    id
  }
}

yfilali:
No, it is something you'll need to handle on your own. See a few code samples to deal with this here: https://github.com/graphql-python/graphene/issues/348

yfilali:
In case you are still looking for a complete solution:
https://yacine.org/2017/02/27/graphqlgraphene-sqlalchemy-and-the-n1-problem/

Cito:
Thank you for that blog post @yfilali, that was really helpful.

Your `optimize_resolve` method works pretty well for me. By using `joinedload` instead of `subqueryload`, I was able to reduce a deeply nested GraphQL query into a single database query. I had to add all paths, however, not only the leaves (simply removed the `if not p.isLeaf` part when assembling the joins).

Another small problem is that in `resolve_related` you need to translate the field names from GraphQL (camel case) to SQLAlchemy (snake case) with `field = to_snake_case(field)`, otherwise it doesn't work with table columns with underscores in them. You can import the `to_snake_case` method from `graphene.utils.str_converters`.

I feel something like this should be really integrated into graphene-sqlalchemy. Or at least it should provide some helper functions which make it easy to implement such optimized querying, with various loading strategies as options. After all, one of the main advantages of GraphQL over REST is that it solves the N+1 problem. This advantage would be greater if it existed also on the level of the database, not only on the network level.

chanind:
Another option is to use a lazy-loader for SQLAlchemy that is more intelligent about bulk loading. We wrote a custom loader that will inspect the DB session and bulk-load all relations on similar models whenever a relation is lazy-loaded. The result should be the equivalent to using `subqueryload` on all relations upfront, but without needing any custom code to do so. The loader we wrote is available here in case it's helpful to anyone else: https://github.com/operator/sqlalchemy_bulk_lazy_loader

flewellyn:
@yfilali Is that code you wrote for optimize_resolve under the same MIT license as the rest of the library?  I would love to make use of it.

yfilali:
@flewellyn absolutely!

mekhami:
@yfilali I can't seem to access your website, is it still up? Would you mind posting the contents here if not?

yfilali:
[...]

PR #395, opened 2023-06-13
Recreate loader if old loader is on different loop

zeptonaut:
Fix for the problem encountered https://github.com/graphql-python/graphene-sqlalchemy/issues/35#issuecomment-1235616457

This change makes it so that the batched data loader is recreated at field resolution time if the current event loop is different than the event loop for the loader.

Previously, the loaders were only created at GraphQL `Schema` creation time. In many systems (including FastAPI and Starlette's `TestClient`), the `Schema` object is required in order to initiate the mount of the GraphQL routes. Thereafter, a new event loop is created (owned by the request handler).

This means that, almost by definition, in many frameworks the initial loaders _will_ be created on one event loop and used on a different event loop in the future. To resolve this problem, this PR makes it so that we only use the memoized loader _if it was created on the same event loop as where the resolution is happening_.

### DIFF
--- a/graphene_sqlalchemy/batching.py
+++ b/graphene_sqlalchemy/batching.py
@@ -137,9 +137,8 @@ def get_batch_resolver(relationship_prop):
             RELATIONSHIP_LOADERS_CACHE[relationship_prop] = loader
         return loader
 
-    loader = _get_loader(relationship_prop)
 
     async def resolve(root, info, **args):
-        return await loader.load(root)
+        return await _get_loader(relationship_prop).load(root)
 
     return resolve
--- a/graphene_sqlalchemy/batching.py
+++ b/graphene_sqlalchemy/batching.py
@@ -137,7 +137,6 @@ def get_batch_resolver(relationship_prop):
             RELATIONSHIP_LOADERS_CACHE[relationship_prop] = loader
         return loader
 
-
     async def resolve(root, info, **args):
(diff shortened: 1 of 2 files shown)


zeptonaut:
Woops - meant to submit this PR on a fork while we test it.

zeptonaut:
@erikwrede any way to get a review on this? Would love to hear if this matches your mental model of how the batching loaders should work.

erikwrede:
Hey @zeptonaut, please excuse the long wait on this issue. The last months have been very busy so this slipped through. The issues with dataloader and event loops are known and a better solution is definitely necessary 😊 

I'll discuss your approach with the others in our discord meeting on nonday @ 9PM CEST and will get back to you.

zeptonaut:
Hi @erikwrede : for what it's worth, this is no longer blocking us as we figured out an alternate way to do data loaders. Ultimately, we ran into problems when we updated to a newer version of Graphene or Graphene SQLAlchemy (can't remember) and this patch started causing problems.

I do think though that the general idea though in this PR holds and that, like you said, figuring out a better mapping between data loaders and event loops probably needs to happen.

erikwrede:
@zeptonaut can you share how you handle dataloading now? Always looking for some new, proven approaches. A QueryBuilder of some sort is definitely on our list for the future, due to the known inefficiencies with dataloading and coroutine state explosion in large queries.

PR #400, opened 2024-01-11
Automatic sort argument for SQLAlchemyInterface

gbunkoczi:
When using SQLAlchemyConnectionField with a SQLAlchemyInterface to get a polymorphic connection, one has to set the sort argument to None to disable automatic sort enum generation. However, the type check in register_sort_enum is too strict, and being an instance of SQLAlchemyBase seems to be sufficient for the code to work.

Tests run with Python 3.8 and 3.12.

### DIFF
--- a/graphene_sqlalchemy/registry.py
+++ b/graphene_sqlalchemy/registry.py
@@ -95,15 +95,10 @@ class Registry(object):
         return self._registry_enums.get(sa_enum)
 
     def register_sort_enum(self, obj_type, sort_enum: Enum):
+        from .types import SQLAlchemyBase
 
-        from .types import SQLAlchemyObjectType
-
-        if not isinstance(obj_type, type) or not issubclass(
-            obj_type, SQLAlchemyObjectType
-        ):
-            raise TypeError(
-                "Expected SQLAlchemyObjectType, but got: {!r}".format(obj_type)
-            )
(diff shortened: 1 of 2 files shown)


gbunkoczi:
Apologies for the delay. I put some tests in place.

Somewhat tangentially, there seems to be more cases of "issubclass(obj_type, SQLAlchemyObjectType)" tests in the codebase, but since these were not triggered so far, I have not changed them. In addition, "- obj_type : SQLAlchemyObjectType" appears multiple times in function docstrings where " -obj_type: SQLALchemyBase" would be more appropriate, e.g.
https://github.com/graphql-python/graphene-sqlalchemy/blob/9c2bc8468f4c88fee2b2f6fd2c0b8725fa9ccee3/graphene_sqlalchemy/enums.py#L94
https://github.com/graphql-python/graphene-sqlalchemy/blob/9c2bc8468f4c88fee2b2f6fd2c0b8725fa9ccee3/graphene_sqlalchemy/enums.py#L166

(these call registry.register_sort_enum, which has been switched to test SQLALchemyBase). I am happy to correct these as part of this PR if this would help.

erikwrede:
@gbunkoczi Awesome! if you have time, please feel free to go ahead. Thanks for your effort! 😊

gbunkoczi:
The checks have all been changed. I tried each call with a SQLAlchemyInterface instance, and all worked as expected. The docstrings were also updated when appropriate.

PR #414, opened 2024-09-13
Feat: make new filters feature optional

skolazbynek:
We have developed our own filtering system on top of `graphene-sqlalchemy` previously and have no need for automatic generation of filters on connections and the currently only clutter up our schema. I haven't found a way to disable them in docs, however, after walking through the code, it seemed to me that the option is there - it's just hardcoded to be always `True`. Was there a reason for that, or maybe an oversight?

This is my first time opening a public pull request, so I'm open to feedback and pointers! Thanks a lot for your work!


erikwrede (approved):
that definitely was an oversight. Thanks for fixing!

PR #420, opened 2025-01-28
Do not create filter class if create_filters is False

romeroyonatan:
### Issue
The Filter classes are being created although `create_filters` is `False`.

For example:
```gql
input PetTypeFilter {
  and: [PetTypeFilter]
  or: [PetTypeFilter]
}
```

### Solution
Check `create_filters` before creating the Filter class.

### Related
https://github.com/graphql-python/graphene-sqlalchemy/pull/414


erikwrede:
Thank you!

